PR #1, opened 2017-07-25
Initial set up with two tests passing

manishavi:


instagram/src/CommentSection.js
```diff
@@ -0,0 +1,25 @@
+import React ,{ Component } from 'react';
+
+class CommentSection extends Component {
+  constructor(props) {
+    super(props);
+
```
ryanhca: extra space here.

instagram/src/CommentSection.js
```diff
@@ -0,0 +1,25 @@
...
+  }
+  ComponentDidMount() {
+    this.setState({
+      comments: this.props.comments
+    });
+  }
+
+  render() {
+    return ({
+      <input>
```
ryanhca: I think if you took a bit of time, you could finish this :)

PR #4, opened 2017-07-27
I forgot to do a pull request

WizrdJim:
...Done


instagram/src/App.js
```diff
@@ -0,0 +1,19 @@
+import React, { Component } from 'react';
+// import logo from './logo.svg';
```
ryanhca: If you're not using this you should probably just delete it.

instagram/src/CommentSection.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+
+class CommentSection extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      input: '',
+      comments: this.props.comments
```
ryanhca: No need for this as you already have access to 'props' because they're being pulled in the constructor.

ryanhca: in fact, it might be better you use componentDidMount to set the comments onto the state of this component.

PR #5, opened 2017-08-03
React-II Sandeep Chandran

deepsun80:
No node modules included in Github.
Used react-bootstrap for styling.


instagram/src/Comments.js
```diff
@@ -0,0 +1,21 @@
+import React, {Component} from 'react';
+import './App.css';
+
+export default class Comments extends Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            comments: {}
```
ryanhca: If this is going to become an array of comments, you should initialize it to an array.

instagram/src/PostInfo.js
```diff
@@ -0,0 +1,38 @@
...
+export default class PostInfo extends Component {
+    constructor(props) {
+        super(props)
+        this.state = {
+            post: {},
+            comments: []
+        }
+    }
+    componentDidMount() {   
+        this.setState({post: this.props.post}),
```
ryanhca: You can combine these two setState functions into one line. 
`this.setState({post: this.props.post, comments: this.props.post.comments});`

PR #6, opened 2017-08-03
Jesse/Drew React-Instagram Clone

drewdiokno:


src/CommentSection.js
```diff
@@ -0,0 +1,51 @@
...
+        e.preventDefault();
+        const text = findDOMNode(this.refs.commentInput).value;
+        const newComment = {
+            username: 'Sean',
+            text
+        };
+        this.setState({
+            comments: [...this.state.comments, newComment]
+        });
+        findDOMNode(this.refs.commentInput).value = '';
```
ryanhca: There are better ways to handle selecting an element from within a react component. By better I just mean more commonly seen today as best practices.

PR #7, opened 2017-08-04
turning in project

aaren-george:


instagram/src/App.js
```diff
@@ -0,0 +1,39 @@
+import React, { Component } from 'react';
+import { PostContainer } from './components/PostContainer';
+import postData from './application-data';
+import LikeButton from './components/LikeButton';
+import NavBar from './components/NavBar';
+import './App.css';
+
+class App extends Component {
+  constructor(props) {
```
seanchen1991: No need to pass `props` to `constructor` and `super` in this case since you aren't calling `props` inside the constructor.

instagram/src/App.js
```diff
@@ -0,0 +1,39 @@
...
+    super(props);
+    this.state = {
+      postData: [],
+      comments: []
+    };
+  }
+
+  componentDidMount() {
```
seanchen1991: Doing this is fine, it won't break anything, but it also isn't actually necessary. You can just hand off the postData into the `PostContainer` component directly without having it be placed in this component's state.

instagram/src/User.js
```diff
@@ -0,0 +1,20 @@
+import React, { Component } from 'react';
+
+export default class User extends Component {
+  constructor(props) {
```
seanchen1991: See comment above about `props`.

instagram/src/User.js
```diff
@@ -0,0 +1,20 @@
...
+    super(props);
+    this.state = {
+      user: {},
+      comment: {}
+    };
+  }
+  componentDidMount() {
```
seanchen1991: Same thing here. There's no actual need to store the user in `this.state`. Inside of the `div` tag, you can just access the user's name with `this.props.user.name`.

instagram/src/components/LikeButton.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import '../App.css';
+
+class LikeButton extends React.Component {
+  constructor(props) {
```
seanchen1991: See above comment about `props`.

PR #9, opened 2017-08-04
Drew and Jesse - completed Instagram clone styling

jranon:
Drew originally forked the repo, but pushes weren't able to get the files inside the /instagram folder.


instagram/src/App.js
```diff
@@ -0,0 +1,38 @@
...
+        postData
+      };
+      this.handleSearch = this.handleSearch.bind(this);
+  }
+
+  componentDidMount() {
+    console.log('loaded and ready to go');
+  }
+
+  handleSearch(event) {
```
seanchen1991: The way you're handling search here won't work I think. You need to split up updating the state that stores the input string and actually performing the submission to clear the input. Here, you're never updating the state, so the search bar will only ever get the input "search".

instagram/src/SearchBar.js
```diff
@@ -0,0 +1,25 @@
...
+
+ const SearchBar = (props) =>  {
+        return (
+            <div className="Header">
+              <div className="Left">
+                <div className="Logo"><i className="fa fa-instagram"></i></div>
+                <div className="Branding">Instagram</div>
+              </div>
+              <div className="Middle">
+                <input type='text'value={props.search} onChange={props.handleSearch} />
```
seanchen1991: The `onChange` handler here would receive your function that updates the state that stores the input, not the function that executes the search itself. With this, you'll be searching on every change to the input.

PR #10, opened 2017-08-04
finished react sprint

p00gz:


instagram/src/CommentSection.js
```diff
@@ -0,0 +1,33 @@
+import React, { Component } from 'react';
+import { Row, Col } from 'react-bootstrap';
+
+export default class CommentSection extends Component {
```
seanchen1991: It's better to move the `export default` syntax to the bottom of the file. Doing that will save you some issues when working with Redux.

instagram/src/PostContainer.js
```diff
@@ -0,0 +1,56 @@
+import React, { Component } from 'react';
+import CommentSection from './CommentSection.js';
+import { Image, Row, Col, Label } from 'react-bootstrap';
+
+export default class PostsContainer extends Component {
```
seanchen1991: See comment above.

instagram/src/SearchBar.js
```diff
@@ -0,0 +1,19 @@
+import React, { Component } from 'react';
+import { FormGroup, FormControl, ControlLabel } from 'react-bootstrap';
+
+export default class Searchbar extends Component {
```
seanchen1991: See comment above.

PR #26, opened 2017-10-26
Dan Volosnikov React-II

lightofdavinci:


lightofdavinci:
added componentDidMount to the comment component

PR #47, opened 2017-12-07
Moises Dobarganes | +Extra Credit Complete | Some Tests Fail

designerexpert:
All content is located inside Instagram/src/components/
Tests were looking for components inside ./src/ , so I redirected them to ./components/
Functionality implemented how tests expected, but tests not functioning properly on my side.
Extended functionality to ability to add comments to posts.
Search Bar collects data and saves it to this.state, but no functions implemented for filtering posts yet.


designerexpert:
Implemented Search Functionality:
(1) Works while typing: Is always one key behind...not sure why..
(2) Works after pressing enter, on search
(3) If nothing is found, displays nothing.
(4) Initially it will display every post
(5) Shows all posts by Username Matching the current Entered Criteria.
   Example: p will display both philzcoffee and playhearthstone.
Made some additions to dummy data for testing.

PR #94, opened 2018-02-06
Peter Gray - Intial Commit - WIP

PeterGrayCreative:


mixelpixel:
I ❤️  the way you are using commit messages! You get a gold ⭐️ !

I downloaded your repository and installed its NodeJS modules. When I launched your web application, I noticed a bunch of error messages (they'll pop up after `npm start`, or in the Chrome Dev Tools Console. Just want to make sure you are aware of them. Not always easy to glean meaningful information from these things, but in this case they are pretty straightforward. And, in general, "Warnings" are not a show stopper like Error messages can be. Page still loads and looks ~great!~ ...alright. 😉  I know you had a lot thrown at you for week 1 of React, but just putting these little details on your radar - everyone loves a message free console when their page loads 😺  

1. `img elements must have an alt prop, either with meaningful text, or an empty string for decorative images` Get in the habit of adding `alt="awesome image description for when the link doesn't load"`
2. `Links must not point to "#". Use a more descriptive href or use a button instead`
3. `Warning: Each child in an array or iterator should have a unique "key" prop.` Have you crossed this bridge yet? In this instance, the warning message in the Chrome DevTools console _actually_ gives you some great information, this link: https://fb.me/react-warning-keys READ IT. Share it with your friends in class! 😛  It is a known pain point, and not at all intuitive. Mostly, React just encourages a "best practice" (and with good reason) and wants something like this when using .map(), even though the key never gets used:
```
        {this.props.users.map((user, i) => {
          return <li key={i}>{user.username}</li>;
        })}
```
4. `Warning: Encountered two children with the same key, `dennis_futbol`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.` See if reading the above doc doesn't give you some thoughts on how to resolve this warning message.

Great work, man! Looks to me like you are doing well with what can be a very difficult subject to get rolling with 👍

PR #104, opened 2018-02-07
<Leon> Work in Progress

LeonOfTheHollow:
--Added directories and files per README instructions
--Implemented React component for search bar @ src/components/SearchBar/SearchBar.js
--Implemented React component for posts @ src/components/PostContainer/PostContainer.js
  --NOTE: this is a stripped-down version of the component for testing, and does not include comments or images.


tsj7:
Hey Leon,

I checked out a few of your files and like you said they look pretty stripped down. That's totally fine if you want to use different versions of files to test different things. In the future by the end of the sprint push up all of your finished files, even if they are not complete push up what you have. That way we can check it out and give feedback, etc.

Tyge

PR #111, opened 2018-02-07
React II finished main goal

germancin:


tsj7:
German,

Nice job on this assignment. I appreciate the cat pic too 😹 . It's well laid out, has no errors in the console and works well. I would recommend committing and pushing more often. This will help organize your projects and be really helpful if you need to revert to a previous commit. Putting helpful commit messages like "searchbar working" also can be helpful.

Nice work.

Tyge

PR #114, opened 2018-02-07
Igor Yermak -WIP

didacta:
directory setup


tsj7:
Hey Igor,

Thanks for submitting a pull req for this assignment. One thing that will help your with your projects is to commit more often and make context specific. Example: make a commit after the `searchBar` is working and give the commit message a title of somthing like `add searchbar component`.  Were you able to work on any of the components? if not that's okay I just didn't see any component folders or files.

thanks,
Tyge

PR #123, opened 2018-03-07
WIP - AD Faris, Joshua Goode

JoshuaGoode:


AndjDavis:
Looks really good, really clean layout, I could follow your logic with ease.   Commits look good.  I could add commits, I could like the post and see the likes increase/decrease.  Time stamped looked good.  This is really good work, good job!

PR #129, opened 2018-03-08
AJ Genung  Jacob Henry  WIP

DragonRydrz:


mvpache:
Missing some functionality, but you are close and the app looks clean. If you have time, work out the add comment stuff. 

The main thing I want to address here is the assignment of state in each child component to the props. This can create a big headache in larger applications. The idea behind state is a single source of truth. Creating a new state can cause problems when you go to update it, because now you have to update all your states so that the app is on the same page. If two components are using the comment portion of the data for example, they can become out of sync quickly if they each have their own state. 

If you go back to refactor, try to refactor the state/props issue as well.

PR #132, opened 2018-03-08
Randy and Courtney

Phoenixrr2113:


mvpache:
Good work, if you have time I'd like you to hit these two things: 
Refactor postContainer to map the comments into the comment component. You'll have three nested components at that point App > Post > Comment. Let me know if you need clarification or help.

Then Get the submit comment up and running. The design pattern will be very similar to the first React project. 

And double EC if you squeeze in the propTypes

PR #146, opened 2018-03-08
SaaSha Pina W.I.P

saashapina:


nomyard:
Hey SaaSha. This was a good attempt. Unfortunately the code that was submitted was the broken code from yesterday before we fixed those bugs. It's good practice to make sure you're submitting code that is not pulling any errors or warnings. 

You didn't get too far with this one. Which is okay! But I would really like to see you tackle this a bit more when you can so you understand react's nuances a bit more. Remember that it's very tough to get this stuff on the first go-around, so it's always a good idea to get those reps in.

nomyard:
Please close this PR when read- thanks!

PR #159, opened 2018-04-04
Eric Gomez React-Insta-Clone

EricGomez101:


DonnieU:
*App.js*: Proper build of class App and its methods. Succinct code and nice use of spread operator in the PostContainer component.

*commentSection.js*: Great use of a class, props, and the required methods of `componentDidMount()` and `render()`. Great `submitComment` method including the use of a conditional for the press of the Enter key to submit a comment. Great touch of adding `onKeyPress`. Good design choice on creating a unique key for each entry using the comment and index.

*postContainer.js*: Nice use of a class for PostContainer. Very good solution on choosing to use React.Fragment to handle the return of multiple elements.

*searchBar.js*: Proper use of props and what is returned.

This is an extremely well done project that's reactive, takes new comments, designed very close to the requirement, and even includes a heart icon that changes color when hovered over! Proper file structure by having a components directory. Good commit messages too! Phenomenal job!

Suggestion: use Pascal case (e.g.- SearchBar.js) for the filenames of your components as you should find that quite common in industry.

Rating (0-3): 3

PR #175, opened 2018-04-04
Subra working with Kaitlyn Flynn & Nalee Riddell 

Sy-CoDe:


justinborek:
## What Went Well

#### Good work on this one guys. The code looks good, and from what I can tell, it looks like the structure is there and the CSS looks right as well. I was unable to test the files, so I don't have a good view of the styling, but from what I can see it looks like solid work so far.

## Areas for Improvement

#### Keep pushing for commiting at LEAST once per hour, with detailed commit messages. It can save you in case something goes wrong on a project. You definitely should have more commits than `initial setup` and `project complete` :laughing:  Also, please make sure you're committing all of the project files and not deleting package.json or any of those types of files. That was the reason I was unable to test code.

## Questions

#### None

## Rating (1-3)

#### 2

PR #180, opened 2018-04-05
Rachelle Pestanas -WIP-

raaayshell:


justinborek:
## What Went Well

#### The code looks good, and from what I can tell, it looks like the basic structure is there. I was unable to test the files, so I don't have a good view of the styling, but nice work getting where you are.

## Areas for Improvement

#### Try to make sure that while you're working, you're commiting at LEAST once per hour, with detailed commit messages. It can save you in case something goes wrong on a project. Also, please make sure you're committing all of the project files and not deleting `package.json` or any of those types of files. That was the reason I was unable to test code. 

## Questions

#### None

## Rating (1-3) 

#### 2

PR #183, opened 2018-04-05
Jacob Lyerla

JacobLeonLyerla:


JacobLeonLyerla:
This can only really be viewed at 800px i need to get the sizing down.

DonnieU:
*App.js*: Proper assignment of `this.state`. Great use of React.Fragment but if the components were wrapped in a <div> I don't think it necessary.

*CommentSection.js*: Proper use of props. Again, was React.Fragment necessary here since you're only returning everything in a wrapped <div>?

*PostContainer.js*: Nice use of a class here. Well done here!

*SearchBar.js*: proper use of props. Code is succinct w/o bloat.

Ample number of commits w/ great commit messages. Excellent job on this project overall!

Rating (0-3): 2.5

PR #193, opened 2018-04-05
Insta clone nearly complete.

ShawnStewart:
I still need to add a header at the top and add new comment functionality.


DonnieU:
*App.js*: Proper use of class, constructor, `this.state`, `this.setState` (componentDidMount()), and nice assignment of `postProps` to `this.state.Posts`.

*CommentList.js*: Nice use of a class here instead of a cb function along w/ proper build-up of the class. Nice to see use of BEM conventions here.

*PostList.js*: Proper use of props and .map. More usage of BEM is good to see. 

Suggestion: Put your components into a `components` directory to follow industry standards.

Great layout w/ sufficient number of commits w/ good commit messages. Keep up the great work!

Rating (0-3): 2.5

PR #198, opened 2018-04-06
Anthony-Calderaro's PR

Anthony-Calderaro:


DonnieU:
*App.js*: Good start to code base. You need a `this.state` in your constructor. Good use of parameters in your PostContainer.

*CommentSection.js*, *PostContainer.js*, *SearchBar.js*: Good template. FYI, you would still need to use `props`. 

Good file structure by placing components into a components directory. Good commit messages.

Still a lot to do but it's a good start to the project.

Rating (0-3): 1.0

PR #225, opened 2018-05-02
Mathias Bakken

MatthiasBakken:


designerexpert:
# Comment
* Great job getting the posts to display properly and overall styling the app. You are effectively mapping each of those posts and passing the props around utilizing very Good naming conventions. 👍 
* Consider adding key props every time you use Map to generate multiple components.
* Continue making meaningful commit messages as you continue working on this project. 👍 
* I look forward to seeing the Comments Section implemented.

designerexpert:
# Comments
* I see you implemented the Input section in the CommentSection component, and it is working to add new comments. Also liked that you used a reactstrap button on that section.
* The search bar disappears from the navbar when I reach around 768px width, not sure if this was intended.
* Overall you demonstrate a good understanding when creating React Components, and Passing Props down from one component and into another. 
* You have a good understanding of the React State object and utilized Component live-cycle methods to load props into the state.
* You are implementing setState() to modify the React State and this is a very good practice, also you are not mutating state indirectly. Great Job! 👍

PR #256, opened 2018-06-18
Libby Hart - cs11 [React-Insta-Clone]

LibbyHart:


bdurb:
@LibbyHart you are on the right track here.  Remember in able to update comments, likes, etc they will have to be on state somewhere.  Think of which components would make the most sense to be stateful, and go from there.  Keep up the good work!

PR #264, opened 2018-06-18
William DiFulvio [React-Insta-Clone]

Wdifulvio523:
Initial commit. Created component folders


metten0:
Nice layout! Looks good. There is a commit that indicates that you created the auth page, but I was not presented with one upon `yarn start`, and I did not see authentication in your components folder. 

I would like to see more commits overall, and I would like you to get into the habit of pretending that you are in a professional production environment and using commit messages that tell the next engineer that checks out your branch exactly what you have accomplished with each push (even after you have gone home for the day). They are not bad, but it is a good thing to think about right before you push up your changes. 

Thanks for your hard work, and feel free to DM me any questions.

Thanks,

Jason

PR #268, opened 2018-06-18
Kevin Tena [React-Insta-Clone]

kevten22:


metten0:
Layout is strong and looks nice. I would appreciate seeing more commits. I also encourage you to pretend that you are working in a professional development environment and writing your commit messages as though you were talking to the next engineer that checks out your branch after you have gone home. They are already pretty good, but it's worth getting in that mindset as early as possible. I know you are going back over some things in pursuit of mastery, but I would like to know if you have made it to auth, and how you might plan to finish up day four MVP. DM me with any questions.

Thanks,

Jason

PR #275, opened 2018-06-18
David Layman React-Insta-Clone

rontou2:


instagram/src/components/postContainer/like.js
```diff
@@ -0,0 +1,43 @@
...
+	this.setState({likes:newvalue})
+  }
+  }
+
+   render() {
+    return (
+      <div className={`${"heartIcon"} + ${this.state.liked}`}>
+        <img onClick={this.handleClick} src={this.state.imgSrc}/>
+      </div>
+    );
```
brcampbell001: Very nice job with getting your like button to work. All you need is for the likes to increment in render and it will be set.

PR #308, opened 2018-06-25
Noah Amar [React-Insta-Clone]

Noah-Amar:


bdurb:
@Noah-Amar I am concerned with your progress on this project.  I am still seeing the boilerplate create react app code when I render your application.  You should be considerably further on this project.  We need to address this.

PR #315, opened 2018-06-25
Yousef/Folder structure completed

Fashoky:


bdurb:
@Fashoky Thank you for the PR, and including your name in the title!

assets/instagram/src/App.js
```diff
@@ -1,7 +1,7 @@
 import React, { Component } from "react";
 import "./App.css";
 import instagram from "./InstagramData";
-import PostsContainer from "./components/PostsContainer/PostsContainer";
```
bdurb: the app is not rendering,  it looks like you don't have the correct path.  Make sure you have your server running and are checking these types of errors.

bdurb:
@Fashoky thank you for the PR.  Your site is not rendering currently due to what appears to be incorrect path names for files.  Make sure you are running your server as you work to address these issues as they arise.

PR #316, opened 2018-06-25
Adalberto May- Setup for project

RockmanExe:


bdurb:
@RockmanExe thank you for the PR!

instagram/src/components/CommentsSection/CommentsSection.js
```diff
@@ -1,12 +1,24 @@
...
+            comments:props.comments
+        };
+}//Constructor
 
-// export default CommentsSection;
\ No newline at end of file
+render() {
+    return (
+        <div className="comments">
+            {this.state.comments.map(comment =>
```
bdurb: 👍

instagram/src/App.js
```diff
@@ -21,14 +22,30 @@ class App extends Component {
     this.setState({ posts:dummyData });
```
bdurb: In order for the filtered posts to render, you will most likely need to set your state inside of a componentDidMount(), this is because it will trigger after render.  otherwise you will likely end up with a blank page at render time.

instagram/src/App.js
```diff
@@ -21,14 +22,30 @@ class App extends Component {
...
       <div className="App">
-        <SearchBar />
+        <SearchBar
+          searchTerm={this.state.searchTerm}
+          searchPosts={this.searchPostsHandler}
+        />
         <PostContainer
-          posts={this.state.posts}
+          posts={
+            this.state.filteredPosts.length > 0
```
bdurb: nice use of the ternary operator to pass props based on the search method.

PR #317, opened 2018-06-25
Kairi Breckenridge React-Insta-Clone

Kaikaine:


bdurb:
@Kaikaine thank you for the PR!

src/instagram/components/Authorize/Authorize.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
```
bdurb: 👍

bdurb:
@Kaikaine looks like you are making some good headway.  Keep it up!

PR #319, opened 2018-06-25
Sania Parekh CS11 React-Insta-Clone

sparekh315:


bdurb:
@sparekh315 thank you for the PR!

instagram/src/App.js
```diff
@@ -1,14 +1,26 @@
 import React, { Component } from 'react';
 import logo from './logo.svg';
 import './App.css';
+import dummyData from './dummy-data';
```
bdurb: @sparekh315 your site is not rendering because the dummy-data file is currently sitting outside of your project.  I would move it inside the src folder, then it will render.

PR #323, opened 2018-06-26
Terrance Danckert CS11 (React-Insta-Clone) Day 1 & 2

Tbarnett20:
***Added Component files***

*******Working on*******
The root App component of your application should import the dummy data from the dummy-data.js file with import dummyData from './dummy-data'; and iterate over said data, passing each individual object as a prop to an instance of PostContainer.


bdurb:
@Tbarnett20 thank you for the pr.

PR #326, opened 2018-07-30
Lauren Worthington

llwor94:


instagram/src/App.js
```diff
@@ -0,0 +1,78 @@
+import React, { Component } from "react";
+import dummyData from "./dummy-data";
+import SearchBar from "./components/SearchBar/SearchBar";
+import PostContainer from "./components/PostContainer/PostContainer";
+import "./App.css";
+
+class App extends Component {
+  state = {
```
ryanhca: Hey! Use the constructor for now, until next week I'll allow you to diverge! :)

instagram/src/App.js
```diff
@@ -0,0 +1,78 @@
...
+    posts: dummyData,
+  };
+
+  handleAddComment = (comment, post) => {
```
ryanhca: This is great! What a flawless execution of this task. If you didn't have to pass the post in, however it might be easier to just add the comment to the comment array found on the CommentSection's state.

There is a bit of a problem here but it's only a slight problem. While passing that post around, you're passing it as a referenced object, which is why your code on line 16 returns true. 

Here's the lesson:

```
const myObj = {name: 'ryan'};
const myObj2 = {name: 'ryan'};
myObj === myObj2  -> returns false. 
These are separate objects not sharing space in memory.
```
example 2 (what you have going on that is a bit problematic);
```
const myObj = {name: 'Lauren'};
const myObj2 = myObj;
myObj === myObj2 -> returns true
These objects share space in memory.
```

So, you're passing around the entire object when this function is invoked (which is totally fine) but then you _could_ potentially create side-effects although you're using the spread operator and essentially `Object.Assign()` and then concatenating the object's array on comments together with the newly added comment... wheeew.. lots to explain and I'm sure you already knew all of this. 

The idea here, and the idea that i gave in the lecture, was to simplify this algorithm by separating our comments away from the actual post object by passing them down as props and setting them on the CommentSection's state.. why does this simply our lives? Well now our algorithm is as simple as.

```
1. build out comment object w/ username, comment and id,
2. clone the state array
3. push comment into clone
4. call setState
```

Quite a few extra steps, however I'm thoroughly impressed with your algorithm.. there are just simply things to consider when it comes to ... simplicity... 

Great stuff

natdf:
@ryanhca has blessed you with reviewing your code! 
I don't have much to add, other than I noticed some inconsistency in your code formatting, which is mostly minor, but a linter may see it as an error. I will point them out to you in different comments. 
Everything looks excellent overall, good job Lauren!

instagram/src/components/App/App.css
```diff
@@ -0,0 +1,15 @@
+/* http://meyerweb.com/eric/tools/css/reset/ 
+   v2.0 | 20110126
+   License: none (public domain)
+*/
+
+
```
natdf: This is mostly a very minor issue, but a linter may see an error when there is more than 1 line space between sections.

instagram/src/components/LogIn/LogIn.css
```diff
@@ -0,0 +1,65 @@
+.login {
+  background-color: #fafafa;
+  height: 100vh;
+  width: 100%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+
```
natdf: Another extra space (more details in other comments).

PR #330, opened 2018-07-30
Sean Kennedy

skennedy93:


instagram/src/login.js
```diff
@@ -0,0 +1,38 @@
+import React, { Component } from 'react';
```
danitacheny: Capitalize the name of this file for consistency, and it'll fix your import in authentication.js (capitalize that one too!)

instagram/src/CommentInput.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+const CommentInput = props => {
+  return (
+    <form onSubmit={this.addComment}>
```
danitacheny: Line 5: It should be `props.addComment`

Line 6: should be `onChange={props.commentHolder}`

PR #332, opened 2018-07-30
Keith Kennedy 

Kennedykid1995:


natdf:
Great work Keith!

- Everything looks good, well done on building the functionality for the "Like" and adding in the comments. 
- You'll notice that you're able to submit empty comments. You might want to see if you can figure out a way to stop that from happening. 
- I like how you styled this, it looks very nice.
- Your code is looking better all the time! Good work. Always have another look through your code to check for consistency. You can run into linting errors if there is more than 1 line space in between sections. 

Excellent job overall!

PR #335, opened 2018-07-30
Francis Tse - React-Insta-Clone

francistse23:
Tasks (Day I)


instagram/src/components/PostContainer/PostContainer.js
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import CommentSection from '../CommentSection/CommentSection';
+import PropTypes from "prop-types";
 
 const PostContainer = props => {
```
dk574: Good job on getting the MVP done Francis! I'm sure this is probably out by now but is this component still in progress? Or did you move it? Just pointing that out but if it's not issue, you can just clean up the rest of your code after you're done. 👍 👍 👍

dk574:
Nice work on the commits Francis! I see you're making progress on your work. Good job! 👍 💯

dk574:
Great job on converting your css to styled components and react-strap! Keep it up. 👍

PR #337, opened 2018-07-30
Justin Hammett | React-Insta-Clone | In-Progress

AI2EJAY:
File structure completed


mikemarshal:
Good start on the project Justin, the login page doesn't seem to load for me right away and the search bar isn't completely functioning. Let me know if you need help settings those up

AI2EJAY:
Is it still like that? I've never had issues with the login page not showing up while a user wasn't signed in. and the search bar works but  uses with includes which is kindof sloppy but I mean it's the same method he used in the solution.

AI2EJAY:
nvm you are right about the login now. I don't know what broke it but i'll look around the code

mikemarshal:
Just tried again... doesn't seem to be. It's odd because the one you had during stand-up was working.. could be something related to localstorage causing the issue

JohnathanHuggett:
Keep up the good work! Everything is looking great for day one. Also, nice use of the textArea for the comment field. Focus on functionally today.

PR #339, opened 2018-07-30
Charles Boegner

MichaelBoegner:
* [ ] Project Setup

* [ ] If you don't have CRA installed then run npx create-react-app instagram to create your starter application with the name instagram.
* [ ] Create a components directory inside the src directory, and then create a sub-directory called SearchBar, another one called PostContainer, and lastly one called CommentSection, all inside the src directory. Each of these directories should contain the component file as well as the CSS for their respective components. You'll also want to put any other components that coincide with your respective container components inside of these directories.
* [ ] Tasks (Day I)
* [ ] At the end of Day I there will be a single instance of the Search Bar being rendered at the top of the page, as well as a Post Container and a Comment Section for every piece of mock data in the dummy-data.js file.
* [ ] The root App component of your application should import the dummy data from the dummy-data.js file with import dummyData from './dummy-data'; and iterate over said data, passing each individual object as a prop to an instance of PostContainer.
* [ ] Each PostContainer component will then pass the array of comments on each post object as a prop to an instance of the CommentSection component.
* [ ] The CommentSection component will receive the array of comments as props and render a Comment component with the username of the poster as well as the post's text. Additionally, there should be an input box that allows users to submit a new comment for any post. We'll work on posting new comments tomorrow.
* [ ] Be sure to check the Types of the data you are passing around as props in the components that will be using props to present data as DOM elements. This should be linked to your Comment component that Comment Section will render and potentially to your Post component that Post Container will render.


JohnathanHuggett:
Very nice commit history! Keep hacking away at that CSS.

instagram/src/components/CommentSection/CommentForm.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+
+
+
+const Comment = props => {
+    return(
+        <form >
+        <input onChange={props.handleComment} placeholder="Add a comment . . ." type="text"/>
+        <button onClick={props.addNewComment}>Add Comment</button>
```
JohnathanHuggett: Move the onClick functions to the form and change it from onClick to onSubmit

PR #342, opened 2018-07-30
Bryce Evans

BryceEvans:


insta-clone/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import './SearchBar.css';
+
+
+const SearchBar = () => {
+    return (
+        <div className="search-bar-top">
+            <div className="logos-wrapper">
+                <i class="fab fa-instagram"></i>
+                <img src = {require("/Users/bryceevans/Desktop/LambdaSchool/React-Insta-Clone/insta-clone/src/assets/insta.png")} alt="logo" className="insta-words" />
```
danitacheny: This line will work as long as the app is running on your local machine, but it needs to use relative pathing for it to work on any other computer!

PR #346, opened 2018-07-30
Charles Cortinas

sherlock270:
Created React App


insta-clone/src/components/PostContainer/PostContainer.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import CommentsSection from '../CommentsSection/CommentsSection';
+import './PostContainer.css';
+import PropTypes from 'prop-types';
+
+
```
natdf: Very minor, but a linter may say there's an error if there is more than 1 line-space between sections.

natdf:
Great job Charles! 

- Everything looks great, and is functioning properly. Well done on building the "Likes" and adding in comments. 
- Your code is on point, though I left another comment on a minor spacing issue. 

Good job overall!

insta-clone/src/components/PostContainer/PostContainer.js
sherlock270: Ok, got it fixed

PR #347, opened 2018-07-30
Jurgen Kela 

kelajatu:


natdf:
Good job Jurgen! 

- The project looks great, well done on getting that functionality built in with the "Likes" and adding comments. 
- Everything appears to be on point with your code, only some minor inconsistency in code formatting, most noticeably near the top just under the imports. I have noticed that even code formatters miss this sometimes for some reason, however some linters may give you errors, so even after using a code formatter it's good to always have a quick look over everything to make sure it's formatted properly. 

Great work overall!

PR #348, opened 2018-07-30
Tim Zulf

Tulf:
### Day I
- There are three major container components that you'll need to implement for this project: the Search Bar, the Post Container, and the Comment Section.
- At the end of Day I there will be a single instance of the Search Bar being rendered at the top of the page, as well as a Post Container and a Comment Section for every piece of mock data in the `dummy-data.js` file.
- The root App component of your application should import the dummy data from the `dummy-data.js` file with `import dummyData from './dummy-data';` and iterate over said data, passing each individual object as a prop to an instance of `PostContainer`.
- Each `PostContainer` component will then pass the array of comments on each post object as a prop to an instance of the `CommentSection` component.
- The `CommentSection` component will receive the array of comments as props and render a `Comment` component with the username of the poster as well as the post's text. Additionally, there should be an input box that allows users to submit a new comment for any post. We'll work on posting new comments tomorrow.
- Be sure to check the `Types` of the data you are passing around as props in the components that will be using props to present data as DOM elements. This should be linked to your `Comment` component that `Comment Section` will render and potentially to your `Post` component that `Post Container` will render.
- You are free to leverage the Bootstrap library for this project for the purposes of theming and styling. I recommend the awesome [reactstrap](https://reactstrap.github.io/) library, which is a library of Bootstrap components that have been implemented using React, so they're really easy to just drop straight into React projects.
- In addition to Bootstrap for theming, you'll want to add your own styles via CSS. To keep things organized, have the CSS file that corresponds with a component live in the same directory as the component file.

### Day II
- **Use lifecycle methods**
- [ ] In `app.js` use `componentDidMount()` to set your data to the component's state. Use the state object to pass data to your `PostContainer` component.

- **Adding comments**

- [ ]   Lets divide up the data that we've been working with this far by separating the comments array onto a new component's state. Pass down the comments through each post to the CommentSection component. Be sure that you set the incoming `comments` props data on the state of the CommentSection component.
- [ ]   Create a function in `CommentSection.js` called `addNewComment` that takes in an event and an index number. The function will add the comment that is on the event object to the post that is at that index number.
- [ ]   Pass the `addNewComment` function down the component tree to where ever you have your 'Add a comment...' input.
- [ ]   The 'Add a comment...' input should be wrapped in a `<form></form>` element. Using that form's `onSubmit` event handler, invoke the `addNewComment` function and pass it the required arguments it needs to add a new comment.
- [ ]   Update your state with the new comment (This should trigger your component tree to "re-render" with your new comment on that post).

- **Liking posts**

- [ ]   This will be set up a lot like the 'Add a comment...' input. Pass a function down to where your heart icon is, and use `onClick` event handler to increment that post's 
- **Search**
- [ ]   Set up the search bar will like the comment input and the like button. In your function, filter out any post whose username doesn't match the search term passed in, then update the state with the resulting data.

### DIFF
--- a/instagram/src/components/SearchBar/SearchBar.js
+++ b/instagram/src/components/SearchBar/SearchBar.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react';
+import React from 'react';
 import './SearchBar.css';
 import  iglogo from "../../assets/iglogo.png";
 import  igicons from "../../assets/igicons.png";
--- a/instagram/src/components/CommentSection/CommentSection.js
+++ b/instagram/src/components/CommentSection/CommentSection.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react';
+import React from 'react';
 import './CommentSection.css';
 import PropTypes from 'prop-types';
 import Comment from './Comment';
--- a/instagram/src/components/Login/Login.js
+++ b/instagram/src/components/Login/Login.js
@@ -1,5 +1,5 @@
 import React, { Component } from "react";
-import { Button, Form, FormGroup, Label, Input, FormText } from "reactstrap";
+import { Button, FormGroup, Input } from "reactstrap";
 import "./Login.css";
 import styled from 'styled-components'
 
--- a/instagram/src/components/PostContainer/PostBody.js
+++ b/instagram/src/components/PostContainer/PostBody.js
@@ -4,11 +4,9 @@ import PostTop from './PostTop';
 import './PostContainer.css';
 import PropTypes from "prop-types"
 import  Likes from './likes'
-import  like from "../../assets/likecons.png";
-import liked from "../../assets/likedcons.png"
 import {
-    Card, CardImg, CardText, CardBody,
-    CardTitle, CardSubtitle, Button} from 'reactstrap';
+    Card,   CardBody,
+  }from 'reactstrap';
 
 class PostBody extends React.Component {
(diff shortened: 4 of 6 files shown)


groov1234:
I see you've setup the file and folder structure, and got some of your imports done. Good job so far.

groov1234:
Looks Great! Missing some functionality though. Are you having trouble with The React Lifecycle `componentDidMount()`? - Make sure anytime you install a new dependency that you also push your package.json. A good practice might be to make a commit after adding a new dependency prior to implementing anything with that dependency.

instagram/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,23 @@
...
+
+    const SearchBar  = (props) => {
+      return(
+        <nav className="navbar navbar-expand-lg navbar-light ">
+          <img alt="instagram logo" src={iglogo} className="logo" />
+
+
+          <div className="row" id="navbarSupportedContent">
+            <form className="form-inline my-2 my-lg-0">
+              <input className="form-control mr-sm-2" type="search" placeholder="Search" aria-label="Search" onkeyDown = {props.searchPosts}></input>
```
groov1234: Hey, Tim. Your `SearchBar` logic is correct. However, there is a typo here on line 14. `onkeydown` should be `onKeyDown`. Great work!

PR #350, opened 2018-07-30
Kyle Meltzer React-Insta-Clone

meltzerkyle:
**Project Setup**
If you so choose you may install create-react-app globally by running yarn global add create-react-app.
If you have CRA installed then run create-react-app instagram to create your starter application with the name instagram.

- [ ]  If you don't have CRA installed then run npx create-react-app instagram to create your starter application with the name instagram.

- [ ]  Create a components directory inside the src directory, and then create a sub-directory called SearchBar, another one called PostContainer, and lastly one called CommentSection, all inside the src directory. Each of these directories should contain the component file as well as the CSS for their respective components. You'll also want to put any other components that coincide with your respective container components inside of these directories.

**Tasks (Day I)**
There are three major container components that you'll need to implement for this project: the Search Bar, the Post Container, and the Comment Section.
- [ ]  Each PostContainer component will then pass the array of comments on each post object as a prop to an instance of the CommentSection component.
- [ ]  The CommentSection component will receive the array of comments as props and render a Comment component with the username of the poster as well as the post's text. Additionally, there should be an input box that allows users to submit a new comment for any post. We'll work on posting new comments tomorrow.
- [ ]  Be sure to check the Types of the data you are passing around as props in the components that will be using props to present data as DOM elements. This should be linked to your Comment component that Comment Section will render and potentially to your Post component that Post Container will render.
- [ ]  In addition to Bootstrap for theming, you'll want to add your own styles via CSS. To keep things organized, have the CSS file that corresponds with a component live in the same directory as the component file.


dk574:
I see your work in the commits! If you have any more updates to push, feel free to do so. Other than that, I'll let you keep at it 👍 👍 👍

dk574:
Nice work on the frequent commits from yesterday. I see you've also started working on Day III tasks as well. When you can, update your task lists above with Day III's tasks. Keep it up! :) 👍 👍 👍

dk574:
Good work on the commit history so far. Update your task list when you get around to it. Keep it up :)

PR #352, opened 2018-07-30
Will Kwon-React-Insta-Clone

wtkwon:
- [x] If you so choose you may install create-react-app globally by running yarn global add create-react-app.
If you have CRA installed then run create-react-app instagram to create your starter application with the name instagram.
If you don't have CRA installed then run npx create-react-app instagram to create your starter application with the name instagram.

- [x] Create a components directory inside the src directory, and then create a sub-directory called SearchBar, another one called PostContainer, and lastly one called CommentSection, all inside the src directory. Each of these directories should contain the component file as well as the CSS for their respective components. You'll also want to put any other components that coincide with your respective container components inside of these directories.


dk574:
Nice work on your progress. In the future, try to add more commits to show different stages of your work. No biggie, but just in case, try to put some spaces in between your text and operators. Your code does look clean though. Good job so far. 👍 👍 👍

dk574:
Nice work on your progress from yesterday! I see that you just got started on today's assignment so I'll let you keep coding away. When you get to it in the end, you can update your tasklist. 👍 💯

dk574:
Your commit history is 👍 👍 👍 Code looks great. You've been doing good with your tasks Will. Keep it up. :)

PR #354, opened 2018-07-30
Jon Simmons CS 12 Insta Clone

Synch3:


instaclone/src/components/commentsection/commentinput.js
```diff
@@ -8,20 +8,20 @@ let CommentInput = props => {
     );
 };
 
-addNewComment = (event, index) => {
```
groov1234: Hey, Jon. You may want to place your `addNewComment` logic inside your `commentSection.js`. Inside your `commentinput.js` is where you will build your form.

PR #357, opened 2018-07-30
Jordan Massingill

jordan-massingill:


instagram/src/App.js
```diff
@@ -3,32 +3,34 @@ import dummyData from './dummy-data';
...
+
   searchItemHandler =(event)=> {
     this.setState({searchInfo: event.target.value});
   }
 
   searchSubmitHandler=(event)=> {
     event.preventDefault();
-    let myPosts= this.state.posts.filter(item => item.username === this.state.searchInfo);
-    console.log(myPosts);
+    let myPosts= this.state.posts.filter(item => item.username.toLowerCase() === this.state.searchInfo.toLowerCase());
```
groov1234: I'm trying to understand the reasoning behind the calls of `toLowercase()` here.

instagram/src/App.js
```diff
@@ -3,32 +3,34 @@ import dummyData from './dummy-data';
...
+    this.setState({posts: dummyData})
+  }
+
   searchItemHandler =(event)=> {
     this.setState({searchInfo: event.target.value});
   }
 
   searchSubmitHandler=(event)=> {
     event.preventDefault();
-    let myPosts= this.state.posts.filter(item => item.username === this.state.searchInfo);
```
groov1234: This was a great start for the search functionality.

instagram/src/App.js
groov1234: after you declared `item` as a function an if statement checking if the username `includes()` the `target` value then return the `item` would work here as well. It works, and is an interesting way of solving the problem, but may not be ideal. Great work!

PR #360, opened 2018-07-30
Naaz Shaikh

noshaikh:


JohnathanHuggett:
Everything is looking amazing! Heading in the right directions and I am pumped of the improvement from last week! Only thing I'll add is to start pushing a little more often. Keep up the hard work.

PR #365, opened 2018-07-30
Steven L

sliw:
##Day 1 


instagram/src/App.js
```diff
@@ -0,0 +1,62 @@
...
+      comment: commentArray,
+      filterData: [],
+      likes: likeArray
+    }
+  }
+
+  handleOnSearch(e) {
+    const input = e.target.value;
+
+    if (e.target.value === '') {
```
danitacheny: I like how you handle these cases.

instagram/src/components/PostContainer/PostContainer.js
```diff
@@ -0,0 +1,109 @@
...
+      filteredData: this.props.filteredData
+    })
+  }
+
+  componentWillReceiveProps(newProps) {
+    this.setState({ filteredData: newProps.filteredData} );
+  }
+  
+  handleLike(e) {
+    const container = e.target.parentNode.parentNode.parentNode;
```
danitacheny: Line 35: This is a little bit unclear. I like the way you're thinking about it, but I wonder if it might be better to have another presentational component that displays the post (maybe called Post?). Then you could pass the index in as a prop, and pass in this method (using either `.bind()` or ES6 arrow syntax) to use within the component as an event listener. Let me know if you need clarification on what I'm saying.

PR #369, opened 2018-07-30
Sania Parekh CS12 React-Insta-Clone

sparekh315:


mikemarshal:
Good progress, but keep pushing those commits so I can see your latest work. Let me know if you need help with tying the components together

PR #371, opened 2018-07-30
Arthur Bates

AliasAJ:
still building components


JohnathanHuggett:
Day one is looking good. I would focus on some styling today then the functionally of adding comments and likes.

PR #374, opened 2018-07-31
React-Insta-Clone-Zachary Baca

cubsfan46350:


instagram/src/App.js
```diff
@@ -0,0 +1,27 @@
...
+import './App.css';
+import dummyData from './dummy-data';
+import Posts from './components/PostContainer/posts';
+class App extends Component {
+  constructor() {
+    super();
+    this.state = {
+      appData: dummyData
+    }
+    dataFunction() = props => {
```
danitacheny: Line 11: This is breaking your code because it needs to be outside the constructor. The syntax is also a bit off:

This line should be `dataFunction = props => {` OR `dataFunction(props) {`

instagram/src/App.js
```diff
@@ -0,0 +1,27 @@
+import React, { Component } from 'react';
+import './App.css';
+import dummyData from './dummy-data';
```
danitacheny: dummy-data.js needs to be inside `src` for this import to work.

instagram/src/components/PostContainer/PostHeader.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import './Posts.css';
```
danitacheny: The file you have is `posts.css`.

For this import to work you have to change this line to import `./posts.css` or rename the CSS file to `Posts.css`.

instagram/src/components/PostContainer/posts.js
```diff
@@ -0,0 +1,22 @@
...
+             thumbnailUrl={props.post.thumbnailUrl}
+            />
+        <div className="post-image-wrapper">
+            <img
+                alt="post thumbnail"
+                className="post-image"
+                src={props.post.imageUrl}
+            />
+        </div>
+        <CommentSection comments={props.post.comments} />
```
danitacheny: You will have to import this. But right now importing it will break it because there's nothing inside the CommentSection file. If I comment out this line, everything else renders fine.

PR #376, opened 2018-07-31
Carey Baldwin

careybaldwin02:


mikemarshal:
The page looks nice Carey! Just a few things I noticed from running the project - search bar isn't functional and comments default to the username 'Harry Potter'. Let me know if you need any help!

PR #377, opened 2018-07-31
James McRae

jaymzsocool:


mikemarshal:
Page looks really nice James, search bar is still a bit wonky. Try starting by having the search bar only return exact queries that match with the data titles (posts)

PR #379, opened 2018-08-06
James Phillips React Tooling MBP

CASDFJ:


instagram/src/components/Comment.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import './CommentSection.css';
+
+const Comment = props => {
+  return (
+    <div className="comment-text">
+      <span className="comment">{props.comment.text}</span>{' '}
```
EricHech: Make sure to look at the image in the README to see what display you're trying to emulate. The username should come before the text, so the order should be changed here. And it might be easiest to just do something like `<div><b>{props.comment.username}</b> {props.comment.text}</div>`.

instagram/src/components/CommentSection.js
```diff
@@ -0,0 +1,52 @@
...
+import PropTypes from "prop-types";
+import Comment from "./Comment";
+import CommentInput from "./CommentInput";
+
+class CommentSection extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      comments: props.comments,
+      comment: ""
```
EricHech: `newComment` might be a better variable name.

instagram/src/components/CommentSection.js
```diff
@@ -0,0 +1,52 @@
...
+  commentSubmit = e => {
+    e.preventDefault();
+    // build out our comment obj
+    const newComment = { text: this.state.comment, username: "Leeroy Jenkins" };
+    // clone our comments array
+    const comments = this.state.comments.slice();
+    // push obj into new clone
+    comments.push(newComment);
+    // set new clone as state... + reset our comment str
+    this.setState({ comment: "", comments });
```
EricHech: Very nice. Experiment with seeing if you can refactor this to use `functional setState` for this, just so that you can learn how to become comfortable with it. If you need help, let me know.

PR #388, opened 2018-08-06
Alex Caves

AlexanderCaves:


instagram/src/components/CommentSection/Comment.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import './Comment.css';
+
+const Comment = props => {
+    return (
+        <div className="comment-text">
+            <span className="comment">{props.comment.text}</span>{' '}
```
EricHech: The order of this should be reversed, right? Username first, text second? Try putting both in a single `div` and then wrap the title in a `b` tag.

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,47 @@
+import React, { Component } from 'react';
+
+class CommentSection extends React.Component {
```
EricHech: So this is cool, I like that you were able to come up with this and do it. However, you're going to have a split in your state. You'll have one set of comments on App-state and another on CommentSection-state, and they're going to diverge. Instead, if you want to keep state here, that's fine, but then only use it for `this.state.comment` and retain the method `commentHandler`. Then move your `submitNewComment` method to your App component and pass it down, and change the logic so that it sets the state on App appropriately. Then, it will re-render everything and drip down so that you'll have a SSOT (single source of truth) on App state. Does this make sense?

instagram/src/components/PostContainer/PostsPage.js
```diff
@@ -0,0 +1,44 @@
...
+  }
+  componentDidMount() {
+    this.setState({ posts: dummyData });
+  }
+
+  searchPostsHandler = event => {
+    const filteredPosts = this.state.posts.filter(posts =>
+    posts.username === event.target.value);
+  
+    this.setState({ filteredPosts: posts });
```
EricHech: If you're declaring `filteredPosts` on line 19, shouldn't you make use of it? Isn't that what you mean to set in the state like this: `this.setState({ filteredPosts: filteredPosts });`? (In which case, you can use the shorthand: `this.setState({ filteredPosts });`.)

instagram/src/components/PostContainer/PostsPage.js
```diff
@@ -0,0 +1,44 @@
...
+      filteredPosts: []
+    };
+  }
+  componentDidMount() {
+    this.setState({ posts: dummyData });
+  }
+
+  searchPostsHandler = event => {
+    const filteredPosts = this.state.posts.filter(posts =>
+    posts.username === event.target.value);
```
EricHech: You need something to contain and build the search value, right? I'm thinking that if you do it this way, and you search for "Eric," it's only going to search for, "E," then, "r," then, "i," and so on. So I think you want to compare it to some value on state like `searchVal` and accumulate the data in there just like we do with input forms. Does that make sense?

PR #389, opened 2018-08-07
first commit- having problems with comment section...other parts are ok

rverdi642:


instagram/src/App.js
```diff
@@ -0,0 +1,28 @@
...
+    this.state = {
+      posts: dummyData
+    };
+  }
+
+
+  render() {
+    return (
+      <div className="App">
+      <SearchBar />
```
EricHech: Fix your indentation here.

instagram/src/components/PostsContainer/PostContainer.js
```diff
@@ -0,0 +1,27 @@
...
+
+const PostsContainer = props => {
+  return (
+    <div className="posts-container-wrapper">
+      {props.posts.map((post, i) => {
+        return (
+          <div>
+            <div>{post.username}</div>
+            <img src={post.imageUrl} alt="post"/>
+            <Post key={i} post={post} />{" "}
```
EricHech: I don't see a Post.js file anywhere in this repo, but you're referencing it here and on line 2.

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+
+
+const Comment = props => {
+  return (
+    <div> 
+    {props.posts.map((username,i) =>  {key={i} username=username )}
```
EricHech: Look carefully at the syntax here.
`{props.posts.map((username,i) =>  {key={i} username=username )}`

It doesn't quite make sense. Let's remove the JSX so that we can read it more easily:

`props.posts.map((username,i) =>  {key={i} username=username )`

Your brackets are mismatched. You're opening a bracket before `key` and before `i`, but you're only closing a bracket after `i`. Also, with what it seemed like you're trying to do, `username` would need to be in brackets as well. I would think that you were intending to do this:

`props.posts.map((username,i) =>  { key={i} username={username} })`

But this doesn't actually make sense. You're acting like you're trying to pass something into a component. So really, you probably meant something like this:

`props.posts.map((username,i) =>  {<SomeComponent key={i} username={username} />})`

But if now you need to either discard those brackets that are opening up the function, or keep them but add in the `return`. So let's discard them so that we can have an implicit return statement:

`props.posts.map((username, i) =>  <SomeComponent key={i} username={username} />)`

Now, this would totally work depending on the surrounding code, and I think it's more in line with what you were trying to do. So, I'm not sure what your intention was, but make sure that you can follow everything I just stepped you through. Get with me if there's any confusion 👍

PR #390, opened 2018-08-31
React-Insta-Clone

m-j-hacker:
@gooseandmegander Tracking


insta-clone/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,70 @@
...
+  addNewComment = (event) => {
+    event.preventDefault();
+    this.setState({
+      comments: [
+        ...this.state.comments,
+        {text: this.state.newComment, username:"THE_USER"}
+      ],
+      newComment: "",
+    });
+
```
gooseandmegander: Right here, you can add `event.target.reset()` to reset the form. :)

PR #397, opened 2018-09-03
Tom Tarpey CS13 - Day 1 MVP and stretch completed

decagondev:
It has been and interesting journey the closest font i could get was a google font it looks close enough for a clone site


decagondev:
well another day done. i got the day 3 mvp and some stretch done. got some extrastyles done too.

PR #401, opened 2018-09-03
Intermediate React: Instagram Clone - Alejandro Krasovsky

alejandrok93:


phillipsdev:
Thanks for the PR!

phillipsdev:
You're doing a great job! I see your search bar is still a work in progress but I can see by your code that you have it laid out.  Also a note, when you create a class component, and you wind up having an empty constructor you can turn that class into a functional component. Unless you have plans to add something into that constructor later for a different part of the assignment 😄

alejandrok93:
Thanks! I added most of them as class components cause I wasn't sure what we were going to do with them and if we were going to need state.. if by the end of the week we don't need to use state on some, I'll make them functional components

instagram/src/components/Authentication/Authenticate.js
```diff
@@ -0,0 +1,25 @@
...
+// import App from "../../App/App";
+
+const authenticate = App =>
+  class App extends React.Component {
+    constructor() {
+      super();
+      this.state = { loggedIn: false };
+    }
+    // componentDidMount() {
+    //   if (localStorage.getItem("username")) {
```
phillipsdev: (I know you have this commented out for a reason) But remember, that loggedIn is part of state and when you change the state you should call setState, but you're on the right track

instagram/src/components/LoginPage/Login.js
```diff
@@ -0,0 +1,44 @@
...
+  constructor() {
+    super();
+    this.state = { inputText: "" };
+  }
+
+  login(event) {
+    event.preventDefault();
+    let username = this.state.inputText;
+    localStorage.setItem("username", username);
+    //location.reload();
```
phillipsdev: If you check out the docs on MDN for Location.reload(), they would give you an example that shows window.location.reload() which is what would work in this situation.

instagram/src/App.js
```diff
@@ -0,0 +1,26 @@
...
+      </div>
+    );
+  }
+}
+
+const HOCApp = Authenticate(App);
+
+return <HOCApp />;
+
+export default App;
```
phillipsdev: For lines 22-26 you're on the right track, however you don't need lines 22-24. In your export you could just say `export default Authenticate(App)`

alejandrok93: what would the Authenticate(App) do in this case? does it export the App component by executing it through Authenticate first every time ??

instagram/src/components/Authentication/Authenticate.js
alejandrok93: oh yeah, that's true, i need to change the state by using setState()..
i have commented it out because I can't get it to render either <App /> or <LoginPage /> still..

PR #419, opened 2018-09-04
Initial submission.

keirankozlowski:
@gooseandmegander


instagram/src/App.js
```diff
@@ -0,0 +1,55 @@
...
+
+  search = event => {
+    event.preventDefault();
+    let additionalPosts = [...this.state.postInfo];
+    additionalPosts = additionalPosts.filter(post => {
+      if(post.username === this.state.search) {
+        return post;
+      }
+    })
+    this.setState({postInfo: additionalPosts})
```
gooseandmegander: Under this line you can add `event.target.reset()` to reset the form.

instagram/src/components/CommentSection/CommentInput.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+function CommentInput(props) {
+    return (
+        <form onSubmit={props.addNewComment} onChange={props.handleChanges}>
+            <input
+              type = "text"
+              name = "comment"
```
gooseandmegander: You have `newComment` on your state in CommentContainer, but you're calling `name="comment"` here in CommentInput. If you change name to `name="newComment"`, your handleChange function will work dynamically, rather than having to call this.setState 2x.

instagram/src/components/CommentSection/CommentContainer.js
```diff
@@ -0,0 +1,79 @@
...
+                {username: "User", text: this.state.newComment}
+            ],
+            newComment: ""
+        });
+        console.log(this.state.comments);
+    }
+
+    handleChanges = event => {
+        this.setState({[event.target.name]: event.target.value})
+        this.setState({newComment: event.target.value});
```
gooseandmegander: You don't need this line: `this.setState({newComment: event.target.value});`. It's basically a hack since you had `name="comment"` on your input element instead of `name="newComment"`. If you make those changes, you should only be using `this.setState({[event.target.name]: event.target.value})` in the handleChanges function since `event.target.name` evaluates to `newComment`.

PR #420, opened 2018-09-04
Andrew Ghobrial --- React-Insta-Clone

ghobs91:
@gooseandmegander


instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import Comment from './Comment';
+import './CommentSection.css';
+
+CommentSection.propTypes = {
```
gooseandmegander: You'll need to import PropTypes into the file to use them.

gooseandmegander: Once you import PropTypes, you'll get a warning that CommentSection is being used before it's called. You can move your propTypes object tot he bottom of the file to silence that warning.

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,52 @@
...
+    postData: PropTypes.shape({
+        comments: PropTypes.array,
+        timestamp: PropTypes.string
+    })
+};
+
+class CommentSection extends React.Component {
+
+    state = {
+        comments: this.props.post.comments,
```
gooseandmegander: In PostContainer, you named your prop containing the comments array 'comments', but you're trying to use `this.props.post` in this file where you didn't pass down a `post` prop. It should be `this.props.comments`.

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,52 @@
...
+        username: ""
```
gooseandmegander: Since we don't have dynamic username's coming in, you could set the username to a default value like "Bob the Builder", that way a username will show up when you add a comment.

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,52 @@
...
+    };
+
+    componentDidMount() {
+        const id = this.props.post.imageUrl;
```
gooseandmegander: What are you using this id reference for?

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,52 @@
...
+        this.addNewComment();
```
gooseandmegander: Do you need to invoke addNewComment in CDM?

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    addNewComment = event => {
+        event.preventDefault();
+        if (this.state.inputText) {
```
gooseandmegander: You're using `this.state.inputText`, but you don't have inputText living on your state.

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,52 @@
...
+            this.setState({
+                users: this.state.users,
```
gooseandmegander: The same is happening here. You're trying to set a property `user` on state but there is not `user` property on the state object. For addNewComment, the function should only manage the logic for adding a new comment. You DO have a comments property on state. 👍

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,52 @@
...
+    handleInput = event => {
+        this.setState({
+            inputText: event.target.value
+        });
+    };
+
+    render() {
+        return (
+        <div className="commentSection">
+        {props.comments.map((comment) => (
```
gooseandmegander: You added the comments to state, so you can use the state instead of props. `this.state.comments`

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,52 @@
...
+            <Comment comment={comment} key={comment.text} />
+        ))}
+        <input value={props.inputText} onChange={props.handleInput} />
```
gooseandmegander: Take a look at <form>

PR #450, opened 2018-09-10
React-Insta-Clone MVP Abdul Ahmad

Xarfo:


instagram/src/App.js
```diff
@@ -0,0 +1,47 @@
+import React, { Component } from 'react';
```
kait-schorr: You can actually import Fragment here alongside Component so you don't have to use React.Fragment in the rest of the file. 

```import React, { Component, Fragment } from 'react';```

File: instagram/src/App.js [hunk 2]
```diff
@@ -0,0 +1,47 @@
...
+
+
+  render() {
+    return (
+      <div className="App">
+      <header className="search-bar">
+                <SearchBar search={this.search}/>
+            </header> 
+
+             <React.Fragment className="App-renderer">
```
kait-schorr: You cannot add a class name to a Fragment. Fragments just exist to help group child elements, but they don't actually render any element to the page. If you need something to have a class, use a div.

instagram/src/components/CommentSection/Comment.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+const Comment = props => {
+    return(
+        <div className='Comment'>
+            <span>{props.comment.username}</span>
+            <span>{props.comment.username}</span>
```
kait-schorr: Why are you displaying the usernames twice?

instagram/src/components/CommentSection/Comment.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+const Comment = props => {
+    return(
+        <div className='Comment'>
```
kait-schorr: Class names are lower case as best practice.

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,61 @@
...
+        super(props);
+        this.state={
+            comments: props.comments,
+            newComment: ''
+        }
+        
+    }
+ 
+ 
+addNewComment = (event,index) => {
```
kait-schorr: If you include a parameter, but never use it, you can just remove it from the function.

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,61 @@
...
+        return (  
+                <div className='commnet-section-instances'>
+                {this.state.comments.map(comment => 
+                <Comment 
+                    key={comment.username} comment={comment}
+                />)}
+                
+                <form onSubmit={this.addNewComment}>
+       
+                        <input onChange={this.handleInput} type="text" placeholder='Add a comment' value={this.commentInput}  />
```
kait-schorr: I think the value here should be this.state.newComment, though maybe you're structuring your solution differently?

instagram/src/components/PostContainer/PostPage.js
```diff
@@ -0,0 +1,24 @@
+import React, { Component } from 'react';
+import PostContainer from './PostContainer'
+
+
+class PostPage  extends Component {
+    constructor(props){
+        super(props);
+        this.state = {
+            ummyDataList: [],
```
kait-schorr: Be very careful with your spelling!

instagram/src/App.js
Xarfo: Got it! As I understand that React>fragment is used so as to not add any new nodes into your DOM.

instagram/src/App.js
Xarfo: Thank you!

instagram/src/components/CommentSection/Comment.js
Xarfo: Will have the extra prop removed.

instagram/src/components/PostContainer/PostPage.js
Xarfo: Gosh, that should be dummyDataList. I should have just used shorter variable names.

PR #454, opened 2018-09-10
michelle ok

Mokagbue:
trying to get the dummydata to read in postcontainer...

### DIFF
--- a/instagram/src/App.js
+++ b/instagram/src/App.js
@@ -34,3 +34,4 @@ class App extends Component {
 }
 
 export default App;
+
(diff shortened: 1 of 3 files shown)


crypto-jones:
Michelle,

I noticed that you're still working on getting comments to work. Any success yet with that?

Some suggestions I have:

- In App.js, I would simplify state to create an empty array inside of that. From there, you can call `componentDidMount()` and bring in the `dummyData` to populate that array, so you can work with that data set across the project.

- For the comments, you should set up state in a `CommentsContainer` component and you can pull in comments via props and work with state for comments there.

-Be careful with your naming conventions. Components in React use PascalCase vs camelCase for naming the files and naming the components within the files themselves.

Please reach out to me soon if you're still having trouble with getting comments to work.

PR #457, opened 2018-09-12
Insta-Clone

raylomeli:
@jeremy-jomes


crypto-jones:
Hi Ray,

It looks like you tagged the wrong person! My Github handle is @crypto-jones. :) 

Next, it looks like you've done some neat things, such as bring in `reactstrap` and made some nice styling on this project. Very cool. 

I do have some questions:

- I noticed in `App.js` and in the `/CommentsSection` directory, you don't have state anywhere. I'm curious how you're dealing with the comments from both `dummyData` as well as the being able to post new comments as there aren't any comment handler functions anywhere.

- Also, I'm not seeing anything regarding HOCs and the login page. Are you okay with those concepts? 

It seems like you're a bit behind. Please reach out if you're blocked anywhere.

PR #462, opened 2018-10-02
Bryce Evans

BryceEvans:
- [x] There are three major container components that you'll need to implement for this project: the Search Bar, the Post Container, and the Comment Section.
- [x] At the end of Day I there will be a single instance of the Search Bar being rendered at the top of the page, as well as a Post Container and a Comment Section for every piece of mock data in the dummy-data.js file.
- [x] The root App component of your application should import the dummy data from the dummy-data.js file with import dummyData from './dummy-data'; and iterate over said data, passing each individual object as a prop to an instance of PostContainer.
- [x] Each PostContainer component will then pass the array of comments on each post object as a prop to an instance of the CommentSection component.
- [x] The CommentSection component will receive the array of comments as props and render a Comment component with the username of the poster as well as the post's text. Additionally, there should be an input box that allows users to submit a new comment for any post. We'll work on posting new comments tomorrow.
- [x] Be sure to check the Types of the data you are passing around as props in the components that will be using props to present data as DOM elements. This should be linked to your Comment component that Comment Section will render and potentially to your Post component that Post Container will render.
- [x] You are free to leverage the Bootstrap library for this project for the purposes of theming and styling. I recommend the awesome reactstrap library, which is a library of Bootstrap components that have been implemented using React, so they're really easy to just drop straight into React projects.
- [x] In addition to Bootstrap for theming, you'll want to add your own styles via CSS. To keep things organized, have the CSS file that corresponds with a component live in the same directory as the component file.

- [x] (Stretch) Implement the ability to comment on a post with the `Add a comment...` input.
- [ ] (Stretch) Use the [moment.js](https://momentjs.com/) library dynamically format the timestamp into a human-readable format like how it is being displayed in the screenshot.

### DIFF
--- a/instagram/src/App.js
+++ b/instagram/src/App.js
@@ -4,10 +4,6 @@ import PostsPage from './components/PostContainer/PostsPage';
 import Authenticate from './components/Authenticate/Authenticate';
 
 class App extends Component {
-  constructor() {
-    super();
-    this.state = {};
-  }
   render() {
     return (
       <div className="App">
--- a/instagram/src/components/CommentSection/CommentInput.js
+++ b/instagram/src/components/CommentSection/CommentInput.js
@@ -20,6 +20,7 @@ const CommentInputForm = styled.input`
     color: grey;
     padding-right: 25px;
     border: white;
+    color: black;
 `;
 
 const Ellipsis = styled.i`
--- a/instagram/src/components/Login/Login.js
+++ b/instagram/src/components/Login/Login.js
@@ -54,7 +54,7 @@ class Login extends Component {
                     <h4>Password</h4>
                     
                     <Input 
-                    type="text" 
+                    type="password" 
                     placeholder="Password..." 
                     name="password"
                     value={this.state.password}
--- a/instagram/src/components/SearchBar/SearchBar.js
+++ b/instagram/src/components/SearchBar/SearchBar.js
@@ -1,7 +1,6 @@
 import React from 'react';
 import logo from '../../assets/instagramlogin.png';
 import styled from 'styled-components';
-import './searchbar.css';
 
 const Header = styled.header`
     display: flex;


BryceEvans:
- [ ] In app.js use componentDidMount() to set your data to the component's state. Use the state object to pass data to your PostContainer component.
- [ ] Adding comments
- [ ] Lets divide up the data that we've been working with this far by separating the comments array onto a new component's state. Pass down the comments through each post to the CommentSection component. Be sure that you set the incoming comments props data on the state of the CommentSection component.
- [ ] Create a function in CommentSection.js called addNewComment that takes in an event and an index number. The function will add the comment that is on the event object to the post that is at that index number.
- [ ] Pass the addNewComment function down the component tree to where ever you have your 'Add a comment...' input.
- [ ] The 'Add a comment...' input should be wrapped in a <form></form> element. Using that form's onSubmit event handler, invoke the addNewComment function and pass it the required arguments it needs to add a new comment.
- [ ] Update your state with the new comment (This should trigger your component tree to "re-render" with your new comment on that post).
- [ ] Liking posts
- [ ] This will be set up a lot like the 'Add a comment...' input. Pass a function down to where your heart icon is, and use onClick event handler to increment that post's likes.
- [ ] Search
- [ ] Set up the search bar will like the comment input and the like button. In your function, filter out any post whose username doesn't match the search term passed in, then update the state with the resulting data.
- [ ] Stretch Problems (Day II)
- [ ] Persist your data using localhost. If done correctly, you will be able to refresh your page and still see your new comments.
- [ ] Add the ability to delete a comment from your data. If your data is in localhost, make sure to delete it from there as well.

instagram/src/components/CommentSection/Comment.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import styled from 'styled-components';
+import PropTypes from 'prop-types';
+
+const CommentText = styled.div`
```
kiadorii: Wow, first time seeing it styled this way!

instagram/src/components/CommentSection/CommentInput.js
```diff
@@ -0,0 +1,46 @@
...
+    color: black;
+`;
+
+const CommentInput = () => {
+    return ( 
+        <div>
+            <InputForm>
+            <InputDisplay>
+                <CommentInputForm type="text" placeholder="Add a comment..." />
+                <Ellipsis className="fas fa-ellipsis-h"></Ellipsis>
```
kiadorii: Hm, from what i can tell, `<Ellipsis>` is not needed here since this is where you are adding your comment. This would be useful for the posts that you may be displaying.

instagram/src/components/CommentSection/Comment.js
BryceEvans: It's called Styled Components. It's covered by Lambda now (I think CS9 didn't get it but CS11 on did). I think it's covered by the lecture next Wednesday!

instagram/src/components/CommentSection/CommentInput.js
BryceEvans: This is a variable that simply refers to an <i></i> html element that is set up above. It is referring to the ellipsis icon of font awesome. Again, it's styled components, which I think will be taught this coming Wednesday. Thanks!

instagram/src/App.js
```diff
@@ -0,0 +1,20 @@
+import React, { Component } from 'react';
+import './App.css';
+import PostsPage from './components/PostContainer/PostsPage';
+import Authenticate from './components/Authenticate/Authenticate';
+
+class App extends Component {
+  constructor() {
```
kiadorii: Should be able to get rid of this since the constructor for a React component is called before it is mounted. Since this is not called before it is mounted, there is no need for it.

instagram/src/App.js
```diff
@@ -0,0 +1,20 @@
...
+    super();
```
kiadorii: Similarly here as well - `super()` calls the constructor of its parent class. This is required when you need to access some variables from the parent class.

instagram/src/components/CommentSection/Comment.js
kiadorii: Speaking of, we did indeed just cover it! 🔥

instagram/src/components/CommentSection/CommentInput.js
```diff
@@ -0,0 +1,40 @@
...
+    border-top: 1px solid lightgrey;
+    width: 97%;
+    height: 40px;
+    text-align: left;
+    font-size: 15px;
+    color: grey;
+    margin: 5px 10px;
+`;
+
+const CommentInputForm = styled.input`
```
kiadorii: Suggestion for comment input to be black when there is something being written. Currently you have it set to the same color as the default grey.

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,61 @@
...
+            comment: '',
+        };
+    }
+
+    commentHandler = event => {
+        this.setState({ comment: event.target.value });
+    };
+
+    commentSubmit = event => {
+        // prevent page refresh
```
kiadorii: Very nice job in commenting here!!

instagram/src/components/CommentSection/LikesSection.js
```diff
@@ -0,0 +1,64 @@
...
+class LikesSection extends Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            likes: props.likes,
+        }
+    };
+
+    incrementLike = () => {
+        let likes = this.state.likes + 1;
```
kiadorii: I believe we covered this in our 1:1 - instagram actually has the increment to be a plus and minus 1. If you were to apply this, how would you go about it?? (Does not have to be in detail, more of a psuedocode thing or just a thought!)

instagram/src/components/CommentSection/LikesSection.js
```diff
@@ -0,0 +1,64 @@
...
+        this.setState({ likes });
+    };
+
+    render() {
+        return ( 
+            <PostFeedbackContain>
```
kiadorii: I really like the naming convention you have - easily understandable.

instagram/src/components/Login/Login.js
```diff
@@ -0,0 +1,80 @@
...
+                    name="username"
+                    value={this.state.username}
+                    onChange={this.inputChange}
+                    />
+                    
+                    <br />
+                    <h4>Password</h4>
+                    
+                    <Input 
+                    type="text"
```
kiadorii:
```suggestion
                    type="password" 
```

instagram/src/App.js
BryceEvans: Correct! It works still!

instagram/src/App.js
BryceEvans: Thanks!

instagram/src/components/CommentSection/CommentInput.js
BryceEvans: I do like that better!

instagram/src/components/Login/Login.js
BryceEvans: Fixed this. Yeah, we talked about doing this and then I never made any changes.

instagram/src/components/CommentSection/LikesSection.js
BryceEvans: I'd have to spend a little bit of time on this, but in theory you would simply use an onClick on the heart icon to toggle a boolean value (true or false) that would be added on state. Then use some logic to increment the count (if true, count + 1; if false, do nothing to the count). Essentially something like this. I may return to tinker with it sometime this week.

PR #463, opened 2018-10-03
Kurt Van Etten

kurtve:
@gakko1  Initial commit


kurtve:
So, no components yet...I just wanted to make sure that I could read from dummy-data.js and throw stuff on the screen.  I'll start decomposing into components tomorrow.

PR #466, opened 2018-10-03
Initial PR

ghost:
@king_atoki

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,14635 @@
+{
+  "name": "instagram",
+  "version": "0.1.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.0.0",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
+      "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
+      "requires": {
+        "@babel/highlight": "^7.0.0"
+      }
+    },
+    "@babel/core": {
+      "version": "7.1.0",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.1.0.tgz",
+      "integrity": "sha512-9EWmD0cQAbcXSc+31RIoYgEHx3KQ2CCSMDBhnXrShWvo45TMw+3/55KVxlhkG53kw9tl87DqINgHDgFVhZJV/Q==",
+      "requires": {
+        "@babel/code-frame": "^7.0.0",
+        "@babel/generator": "^7.0.0",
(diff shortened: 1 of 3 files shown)


src/App.js
```diff
@@ -0,0 +1,32 @@
...
+	}
+
+
+ render() {
+    return (
+      <div className="App">
+			<SearchBar/>
+			<CommentSection/>
+			{dummyData.map( (post, index) => (<PostContainer userPost={post} key={post.username + index} />))}
+			<div>{console.log(dummyData)}</div>
```
KingAtoki: If you want to `console.log` something in the `render` method, you don't have to do it that way, you can just do it outside of the `return` statement and you don't need to wrap it in a `div`

src/components/PostContainer/PostContainer.js
```diff
@@ -0,0 +1,28 @@
+import React, {Component} from 'react';
```
KingAtoki: If you are just creating a functional component, you don't have to import `{Component}`

src/components/PostContainer/PostContainer.js
```diff
@@ -0,0 +1,28 @@
...
+import CommentSection from '../CommentSection/CommentSection.js';
+import dummyData from '../dummy-data.js';
+
+const PostContainer = (props) => {
+	return(
+		<div className="Container">
+			<div className="Wrapper">
+				<img src={dummyData.thumbnailUrl}/>
```
KingAtoki: Don't forget your `alt` attribute on your img

src/components/PostContainer/PostContainer.js
```diff
@@ -0,0 +1,28 @@
+import React, {Component} from 'react';
+import CommentSection from '../CommentSection/CommentSection.js';
+import dummyData from '../dummy-data.js';
```
KingAtoki: You don't have to bring in dummy data because you are passing down the data you need through props

src/components/PostContainer/PostContainer.js
```diff
@@ -0,0 +1,28 @@
...
+
+const PostContainer = (props) => {
+	return(
+		<div className="Container">
+			<div className="Wrapper">
+				<img src={dummyData.thumbnailUrl}/>
+				{dummyData.username}
+			</div>
+
+			<img className="Post-Img" src={dummyData.imageUrl}/>
```
KingAtoki: Because you passed down props properly, you can actually do something like `src={props.post.imageUrl}`

src/components/PostContainer/PostContainer.js
```diff
@@ -0,0 +1,28 @@
...
+
+			<div className="Post-Buttons">
+				<i className="Heart"></i>
+				<i className="Comment"></i>
+			</div>
+
+			<div className="Likes">
+				{dummyData.likes} likes
+			</div>
+			<CommentSection conent={dummyData.comments}/>
```
KingAtoki: I think you have a typo here

KingAtoki:
Don't forget to add PropTypes and defaultProps when necessary

PR #471, opened 2018-10-03
William Yturralde-React-Insta-Clone

willieino:
@raymondgrc

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,8 @@
 
 - The purpose of this project is to continue building on your knowledge of React that you have gained thus far.
 
-  - Here, you'll be implementing a lot of the same concepts that you have been in previous projects, in a very similar fashion.
+  - 
+  Here, you'll be implementing a lot of the same concepts that you have been in previous projects, in a very similar fashion.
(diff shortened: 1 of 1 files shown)


instagram/src/components/PostContainer/PostContainer.js
```diff
@@ -0,0 +1,63 @@
...
+            <input
+              className={props.type === 'action' ? 'addComment2 visible' : 'addComment2'}
+              placeholder="Add comment..."
+              type="text"
+              id="inp2"
+              name="username"
+              value={props.value}
+              />
+           </form>
+          <a className="button" id="commentDots" href="./index.html" onClick={props.postIcon1}></a>
```
Raymondgrc: I would be careful with this href. It seems to me that this would lead you to whatever location you're at + /index.html

instagram/src/components/CommentSection/Comment.js
```diff
@@ -1,22 +1,40 @@
...
+            counter: 0,
+            asyncMessage: ''
+        };
 
-        </div>
+    }
 
-    );
-};
+    increment = event => {
```
Raymondgrc: Don't forget to add a decrement function. It shouldn't be too different.

PR #472, opened 2018-10-03
Julie-Jonak: React-Insta-Clone

juliejonak:
@katiagilligan888 Opening the PR to begin project

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 - The purpose of this project is to continue building on your knowledge of React that you have gained thus far.
 
-  - Here, you'll be implementing a lot of the same concepts that you have been in previous projects, in a very similar fashion.
+  x Here, you'll be implementing a lot of the same concepts that you have been in previous projects, in a very similar fashion.
   - The main difference this time around is that you'll be using the `create-react-app (CRA)` utility to generate your React project for you.
   - Don't worry about all of the extra files that you may not understand which will be present inside your React application that you will build with CRA.
   - The overall structure of the project remains exactly the same as what you worked with in Todo-React.


instagram/src/CommentSection/Comment.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+const Comment = (props) => {
+    return (
+        <div>
```
katiagilligan888: You can also use fragments now `<>` instead of a div to wrap child components in React

instagram/src/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,32 @@
...
+                        <Comment key={item.text} username={item.username} text={item.text} />
+                    )
+                })}
+
+                <input></input>
+
+            </div>
+        )
+}
+
```
katiagilligan888: Try to add Prop Types to all places where props are passing down data

instagram/src/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,32 @@
...
+export default CommentSection;
+
+
+// Icons heart and CommentSection
```
katiagilligan888: Good job pseudocoding the structure of the section before coding out!

instagram/src/CommentSection/CommentSection.js
juliejonak: I added Prop Types to just the PostContainer.js because it checks all through the data, including the comment array. However, I can add additional Prop Type checking in the children functional components if you recommend it?

katiagilligan888: Your prop types look good! Since you type checked all the data including the comments, you are fine to continue. Sorry, I reviewed this before I saw you add the propTypes. Good job!

instagram/src/App.js
```diff
@@ -0,0 +1,62 @@
...
+    })
+  }
+
+  render() {
+    return (
+      <div className="App">
+          <SearchBar data={this.state.data}/>
+          {this.state.data.map(item =>{
+            return(
+            <PostContainer obj={item} key={item.timestamp} />
```
katiagilligan888: Try to use naming that is more specific to what is being passed down in props to the children components, such as userData, instead of passing obj

instagram/src/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,75 @@
...
+        this.setState({
+            comments: [...this.state.comments, newComment],
+            input: ''
+        })
+    }
+
+    addNewLike = (event) => {
+        event.preventDefault();
+        const hearts = this.state.counter;
+        this.setState({
```
katiagilligan888: Good job implementing the like functionality! Currently it continues to increase every time it is clicked on. How would you implement it for a user, without allowing them to like it more than once? Also, here's another way to increment state without setting it to a variable 
` this.setState(prevState =>{
            return {
                likes: prevState.likes + 1, 
}`

instagram/src/Authentication/Authenticate.js
```diff
@@ -0,0 +1,37 @@
...
+        componentDidMount(){
+            if (localStorage.getItem('username')){
+                this.setState({
+                    loggedIn: true
+                })
+            }
+        }
+        
+        render(){
+            if (this.state.loggedIn === false) {
```
katiagilligan888: You could cut this down to only one line of code by using a ternary statement {this.state.loggedIn ? <ConditionalComponent /> :  <PassedComponent />}

instagram/src/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,158 @@
...
+`;
+
+const CommentIcons = styled.div`
+    display: flex;
+    justify-content: flex-start;
+    margin-left: 3%;
+    margin-top: 10px;
+`;
+
+const Icons = styled.i`
```
katiagilligan888: Good job adding styled components. They all look good

instagram/src/PostContainer/PostContainer.js
```diff
@@ -0,0 +1,54 @@
...
+    margin: 0 auto;
+`;
+
+const UsernameHeader = styled.div`
+    display: flex;
+    justify-content: flex-start;
+    align-items: center;
+    margin-bottom: 10px;
+`;
+
```
katiagilligan888: If you are using Styled Components for all of your components and files I would suggest putting all of them in a separate file and then importing them in.

instagram/src/SearchBar/Icons.js
```diff
@@ -0,0 +1,50 @@
...
+    }
+`;
+
+class Icons extends React.Component {
+    constructor(){
+        super();
+        this.state={}
+    }
+
+    logout = (e) => {
```
katiagilligan888: Good job implementing a log out!

PR #474, opened 2018-10-03
Andres Rivera

andresriveratoro:
@47analogy


47analogy:
Console is giving a warning about using a unique key when mapping in App.js and CommentSection.js. Otherwise everything else looks great, with components are set up and passing props properly (PropTypes are checking valid), with the screen rendering as expected.

PR #475, opened 2018-10-03
Abdiel Fernandez

CBMR:
@KingAtoki


CBMR:
I still need to style the whole page

instagram/src/App.js
```diff
@@ -0,0 +1,22 @@
...
+class App extends Component {
+
+  state = {
+    data: data
+  }
+  render() {
+    return (
+      <div className="App">
+        <Search />
+        <Posts data={this.state.data}/>
```
KingAtoki: If you have state just for the sake of passing it to a component you can pass it to the component directly. For example

`<Posts data={data}/>` and the `data` inside of the `{}` is from `./dummy-data`

instagram/src/components/CommentSection/Comment.js
```diff
@@ -0,0 +1,16 @@
...
+
+const Comment = props => {
+  return (
+    <div className='comments'>
+      <p><span className='username'>{props.comment.username}</span> <span>{props.comment.text} </span></p>
+    </div>
+  )
+}
+
+Comment.propTypes = {
```
KingAtoki: If your proptypes here aren't `.isRequired` then you will need to specify defaultProps

instagram/src/components/PostContainer/PostList.js
```diff
@@ -0,0 +1,27 @@
...
+            <p className='timeStamp'>{post.timestamp}</p>
+            <input className='addComment' type='text' placeholder='Add a comment...' />
+          </div>
+        )})}
+    </div>
+  )
+}
+
+PostList.propTypes = {
+ timestamp: PropTypes.string,
```
KingAtoki: _Technically_ timestamp isn't a prop that is being passed directly to PostList, only data is it looks like, so define proptypes for that

instagram/src/App.js
CBMR: I did not know that! Awesome! Thanks

instagram/src/components/PostContainer/PostList.js
CBMR: I don't fully understand the second part.

KingAtoki: You should do something like

```
PostList.propTypes = {
 data: PropTypes.arrayOf(PropTypes.object),
}
```

instagram/src/App.js
```diff
@@ -0,0 +1,38 @@
...
+
+class App extends Component {
+
+  state = {
+    data: [],
+    filter: [],
+  }
+
+  componentDidMount() {
+    this.setState({data: data})
```
KingAtoki: With ES6 if the `key` and the `value` are the same you can shorten. For example 
```this.setState({ data: data }) === this.setState({ data })```

File: instagram/src/App.js [hunk 5]
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  filteredSearchHandler = e => {
+    const posts = this.state.data.filter( post => {
```
KingAtoki: This can be shortened a bit.

```const posts = this.state.data.filter(post => post.username.includes(e.target.value))```

PR #476, opened 2018-10-03
Abi Franklin Insta-Clone

AbiFranklin:
@gakko1


instagram/src/components/Login/LoginPage.js
```diff
@@ -0,0 +1,53 @@
...
+class LoginPage extends React.Component {
+    constructor(props) {
+      super(props);
+      this.state = {
+        username: '',
+        password: ''
+      };
+    }
+
+    handleChange = e => {
```
gakko1: This one is good and should be how your adding comments works as well

instagram/src/components/PostContainer/PostsPage.js
```diff
@@ -0,0 +1,52 @@
...
+      return key.username == e.target.value
+    })
+
+    if (dummyData.length > 0){
+      console.log(dummyData)
+      this.setState({
+      data: {dummyData}
+    })
+  } else {
+    this.allResults();
```
gakko1: Good workaround for showing all results again

PR #478, opened 2018-10-03
Kyle Baker

kybak:


KingAtoki:
For PropTypes if you have somethat aren't `.isRequired` then it would be best to define some default props as well

PR #480, opened 2018-10-03
Rob Salzberg - React-Insta-Clone

ghost:
@juliankohlman

- [x] Setup project

**Tasks (Day I)**

- [x] At  the end of Day I there will be a single instance of the Search Bar being rendered at the top of the page, as well as a Post Container and a Comment Section for every piece of mock data in the dummy-data.js file.

- [x] The root App component of your application should import the dummy data from the dummy-data.js file with import dummyData from './dummy-data'; and iterate over said data, passing each individual object as a prop to an instance of PostContainer.

- [x] Each PostContainer component will then pass the array of comments on each post object as a prop to an instance of the CommentSection component.

- [x] The CommentSection component will receive the array of comments as props and render a Comment component with the username of the poster as well as the post's text. Additionally, there should be an input box that allows users to submit a new comment for any post. We'll work on posting new comments tomorrow.

- [x] Be sure to check the Types of the data you are passing around as props in the components that will be using props to present data as DOM elements. This should be linked to your Comment component that Comment Section will render and potentially to your Post component that Post Container will render.

- [ ]  In addition to Bootstrap for theming, you'll want to add your own styles via CSS. To keep things organized, have the CSS file that corresponds with a component live in the same directory as the component file.

Stretch Problems (Day I)


robsalzberg:
@juliankohlman

- [ ] In  app.js use componentDidMount() to set your data to the component's state. Use the state object to pass data to your PostContainer component.

- [ ] Lets  divide up the data that we've been working with this far by separating the comments array onto a new component's state. Pass down the comments through each post to the CommentSection component. Be sure that you set the incoming comments props data on the state of the CommentSection component.

- [ ] Create  a function in CommentSection.js called addNewComment that takes in an event and an index number. The function will add the comment that is on the event object to the post that is at that index number.

- [ ] Pass  the addNewComment function down the component tree to where ever you have your 'Add a comment...' input.

- [ ] The  'Add a comment...' input should be wrapped in a <form></form> element. Using that form's onSubmit event handler, invoke the addNewComment function and pass it the required arguments it needs to add a new comment.

- [ ] Update  your state with the new comment (This should trigger your component tree to "re-render" with your new comment on that post).

- [ ] This  will be set up a lot like the 'Add a comment...' input. Pass a function down to where your heart icon is, and use onClick event handler to increment that post's likes.

- [ ] Set  up the search bar will like the comment input and the like button. In your function, filter out any post whose username doesn't match the search term passed in, then update the state with the resulting data.

Stretch Problems (Day II)

- [ ] Persist  your data using localhost. If done correctly, you will be able to refresh your page and still see your new comments.

- [ ] Add  the ability to delete a comment from your data. If your data is in localhost, make sure to delete it from there as well.

instagram/src/App.js
```diff
@@ -0,0 +1,36 @@
...
+			data: []
+		};
+	}
+
+	componentDidMount() {
+		this.setState({ data: dummyData });
+	}
+
+	filterPosts = str => {
+		const filteredPosts = dummyData.filter(post => post.username.includes(str));
```
juliankohlman: Nice job making your change to posts outside of your call to `setState()`. Makes the code clean and easy to read/debug 👍

instagram/src/components/CommentSection/commentSection.js
```diff
@@ -0,0 +1,77 @@
...
+		this.setState({ likes: this.state.likes + 1 });
+	};
+	render() {
+		return (
+			<div className="comments-section">
+				<img
+					onClick={this.likePost}
+					className="loveheart"
+					src={require('../assets/loveheart.png')}
+					alt={'loveheart'}
```
juliankohlman: Alt tags/classNames don't need curly braces. Unless you want to evaluate some JavaScript you can omit the curly braces.

instagram/src/components/CommentSection/commentSection.js
```diff
@@ -0,0 +1,77 @@
...
+				/>
+				<img
```
juliankohlman: I'd consider reworking the way a comment gets added here. Maybe a `form` with an `input` element? From the user's perspective, it's not clear that they need to click on the comment bubble to submit their comment after they enter it into the field.

instagram/src/components/CommentSection/Comment.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Comment = props => {
+	return (
+		<div className='comment-section'>
+			<strong>{props.comment.username}</strong>{props.comment.text}
```
juliankohlman: Each `username` is running right up against the `comments text`. Adding a space or some other separator after the username would clean that up.

instagram/src/components/PostContainer/postContainer.js
```diff
@@ -0,0 +1,24 @@
...
+import CommentSection from '../CommentSection/commentSection';
+import './postcontainer.css';
+import Post from '../Post/Post';
+import PropTypes from 'prop-types';
+
+const PostContainer = props => {
+	return (
+		<div className="post-container">
+			{props.posts.map(post => (
+				<div className="post-tunnel" key={post.timestamp}>
```
juliankohlman: 👍 for using the timestamp of each post as it's key

robsalzberg:
@juliankohlman
 
**Tasks (Day III)**

- [ ] Create  a <PostsPage /> component in your components/PostsContainer directory.

- You'll  have to move a lot of what is rendered in app.js to this new component
In app.js, render the PostsPage component.

- Make  sure the app working as it was before since it has been re-factored now.
This is to ensure that we clean up our App component a little bit before we re-factor it to be wrapped up in an HOC

- [ ] Building the High Order Component

- Create  a directory called Authentication
- Inside  that directory create a component called <Authenticate />. This is where all of the magic sis going to happen.
- This  component should be able to take in a component as an argument, and it will return a class component.
-  Inside  of <Authenticate />'s render method, you'll want to return the Component that gets passed into it.
- Be  sure to export out this component.
- Head  over to App.js and import in our new Authenticate Higher Order Component and pass in App.
- If this worked right then everything should render as it used to.
- Authenticate  will look a lot like this when you're done setting it up.
const Authenticate = App =>
  class extends React.Component {
    render() {
      return <App />;
    }
  };

 [ ] Build out the LoginPage component. You can design it how you like

- Inside of Authenticate we need to add a constructor to hold our state data.
- On state we need a user loggedIn boolean flag.
- On componentDidMount we need to check localStorage to see if a user is logged in.
- Inside of the render function we will check if a user is logged in
- If a user is logged in we will return the <App />, else we will return the <LoginPage>

robsalzberg:
Tasks (Day IV)

- [ ] Re-factor  SearchBar to use styled-components.

- [ ] Create   a Header styled-component.header that wraps your entire SearchBar.

- We  recommend working left to right so begin changing out your styles on the instagram logo.
- Create  a LogoHeader styled-component.div
- Create  a LogoImage styled-component.img that fills in the background of your

- [ ] Re-factor  your PostContainer component use only styled components
- Start  with the UserThumbnail and Username
- Note that the Username styles could most certainly be used by your CommentSection component as well.
- To  achieve this, you'll simply just build out your reusable styled-component inside of a Styles/Reusables directory. This is where you could store all of your reusable styled components and export them out of their respective files to then import them for reuse.
- Notice  that the Username on top of the Post is a bit bolder than the Username's found in the comment section. If there is a way to make this reusable styled component accept a prop that will distinguish it from being now is the time to figure that out.

- [ ] Now  move onto the CommmentSection and get rid of any CSS that you don't need.

 - [ ] By now you're a pro at using Styled-Components and we hope you armed with the ammo necessary to choose whether or not you like using it as opposed to vanilla CSS or even a Pre-Processor. We don't want you to believe that there is never a time and place for both native CSS/Pre-processing to exist. But at least now you have a variety of weapons you can choose from to get the job done.

- [ ] Now   is the time to take a step back and look at what you have accomplished this week. Start from the Day 1 README file and see what the tasks were there? Think about how we asked you to approach this problem. Think about the different components you have built? Did you separate out your Comments array from the rest of the application state? If so, what types of advantages did you gain by doing so? React is a very powerful tool and the next steps in learning about the React Ecosystem will all revolve around other libraries that we will plug into our application for extending it's use.

- [ ] Read this article about 'State Driven Views' and come up with a short paragraph as to why you feel that state-driven views are important to us as software developers today? What are state-driven applications? Why are they powerful? How does React enables those patterns?

PR #481, opened 2018-10-03
Alex Martin PR

ghost:
@KingAtoki


KingAtoki:
I see a lot of boilerplate code. Push up your changes so i can see the updated versions of them please

KingAtoki:
Changes are looking good

instagram/src/PostsPage.js
```diff
@@ -0,0 +1,54 @@
...
+class PostsPage extends React.Component {
+	constructor() {
+		super();
+		this.state = {
+			data: [],
+		};
+	}
+	
+	componentDidMount() {
+		this.setState({ data: data});
```
KingAtoki: With ES6 if the key and the value are the same, you can shorten in

For example `this.setState({ data: data});` can be shortened to `this.setState({ data });`

PR #483, opened 2018-10-03
Steve Kutz

stevekutz:
katiagilligan888


instagram/src/components/PostContainer/PostFooter.js
```diff
@@ -0,0 +1,41 @@
...
+    super(props);
+    this.state = {
+      likes: props.likes
+    }
+
+  }
+
+  addLikes = () => {
+    this.setState({
+      likes: this.state.likes + 1
```
katiagilligan888: Do not directly mutate state. On line 16 it would be better to use `this.setState(prevState` =>{
likes: prevState.likes + 1})`

instagram/src/components/PostContainer/PostFooter.js
```diff
@@ -0,0 +1,41 @@
...
+    })
```
katiagilligan888: Also with the like functionality, your counter is currently increasing every time it is clicked. This means that the user can like the post more than once. How would you implement a like that will only increase 1 when clicked, then decrease when clicked again.

instagram/src/components/PostContainer/PostHeader.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import './PostContainer.css';
+
+
+export default class PostHeader extends React.Component {
```
katiagilligan888: This can be a functional component, instead of class

instagram/src/components/PostContainer/PostImage.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import './PostContainer.css';
+
+
+export default class PostImage extends React.Component {
```
katiagilligan888: Also make this a functional component

instagram/src/components/Authentication/Authenticate.js
```diff
@@ -0,0 +1,32 @@
...
+
+const Authenticate = PassedComponent =>
+  class extends React.Component {
+    constructor(props) {
+      super(props);
+      this.state = {loggedIN: false}
+    }
+
+    componentDidMount() {
+      if (localStorage.getItem('user' === 'ImaHacker') &&
```
katiagilligan888: This is the part of your Authentication that is breaking. If you only if a user key exists in the localStorage instead of checking the value, it will let you login. Also on line 15 you have a misplaced parenthesis right after password that should wrap the whole statement

instagram/src/components/Authentication/Authenticate.js
```diff
@@ -0,0 +1,32 @@
...
+      ) {
+        this.setState({loggedIN: true})
+      }
+      else {this.setState({loggedIN: false})
+      }
+
+    }
+
+    render() {
+      if(this.state.loggedIN) return <PassedComponent/>;
```
katiagilligan888: I suggest using a ternary statement here instead of an if statement. `this.state.loggedIN ? <PassedComponent /> : <Login />`

PR #485, opened 2018-10-03
Lee Lichtsinn / React-Insta-Clone

leelichtsinn:
added components directory
added SearchBar, PostContainer and CommentSection directories
@47analogy


leelichtsinn:
Added Post.js and Post functional component
Added PostHeader functional component
Added import for PostContainer and passed state as props to PostContainer
Added CommentSection functional component
Added import for CommentSection into Post.js
@47analogy

leelichtsinn:
added comment-section branch to work out props issue
changed CommentSection component from functional to stateful
passed props as state to CommentSection component
refactored map function
@47analogy

47analogy:
Should be using PropTypes for those components where props are being passed in (i.e. PostContainer.js). Also might be a good idea to place the PostHeader component into it's own file and import it into the Post component when you finish/if have time to refactor. Otherwise, everything is looking great on the screen.

PR #486, opened 2018-10-03
Jason Hedrick FSWPT2

korghal:
@camilamdaniels


instagram/src/Components/Authentication/Authenticate.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+
+
+const Authenticate = (App, Login) => {
+    return class WrappedComponent extends React.Component {
```
TERR1E: You can clean up your code here a bit by just doing
```js
const Authenticate = (App, Login) =>
  class extends React.Component {
 ...statement of codes
}

File: instagram/src/App.js [hunk 2]
```diff
@@ -0,0 +1,117 @@
...
+        searchTerm={this.searchTerm}
+        />      
+      </div>
+    );
+  }
+}
+
+
+
+export default App;
```
TERR1E: Try moving all of the login handlers and functionalities to Login component and keep the app to render the main container only.
Also try to utilize the state more, I see that you are defining the username, comment, and search term outside of the state. Learning how to utilize the state is the number one important thing you have to get a good grasp of.

TERR1E:
Hi Jason,
Here are my suggestions after reviewing your code:

Try moving all of the login handlers and functionalities to Login component and keep the app to render the main container only.
Also try to utilize the state more, I see that you are defining the username, comment, and search term outside of the state. Learning how to utilize the state is the number one important thing you have to get a good grasp of.

I would also suggest moving all the styled components into a separate directory called `styles` and keep it separately from the functionalities

Overall you are doing a great job! 👍

PR #491, opened 2018-10-03
Jesse Anderson - React Insta Clone

Jesse9009:
@juliankohlman


instagram/src/components/Post/PostContainer.js
```diff
@@ -0,0 +1,33 @@
...
+import './post.css';
+
+class PostContainer extends React.Component {
+  render() {
+    // console.log(this.props.data);
+    return (
+      <div>
+        {this.props.data.map(x => {
+          return (
+            <div className="post" key={x.timestamp}>
```
juliankohlman: Using the timestamp for uniqueness on your posts 👍

instagram/src/components/Search/SearchBarLogo.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+class SearchBarLogo extends React.Component {
+  render() {
+    return (
+      <div className="sbLogo">
+        <i className="fab fa-instagram" />
+        <hr />
+        <i className="fab fa-hooli" />
```
juliankohlman: 🤣 Aviato > Hooli

instagram/src/App.js
```diff
@@ -0,0 +1,59 @@
...
+        })
+      });
+    } else {
+      this.componentDidMount();
+    }
+  };
+
+  render() {
+    // console.log(this.state);
+    if (!this.state.dummyData) {
```
juliankohlman: Nice job using a loading message while the user waits for the app to load. You'll use this concept a bunch especially when we get into working with more 3rd party APIs or when we have a backend that we're retrieving data from 👍

instagram/src/App.js
```diff
@@ -0,0 +1,59 @@
...
+
+  handleSearchInput = event => {
+    this.setState({ searchInput: event.target.value });
+  };
+
+  searchPosts = event => {
+    event.preventDefault();
+    if (this.state.searchInput) {
+      this.setState({
+        dummyData: this.state.dummyData.filter(post => {
```
juliankohlman: You could move your call to filter into a separate variable outside of your `setState` call for easier debugging. That way your call to setState can look something like below.
```javascript
this.setState({dummyData: filteredPosts})
```

PR #497, opened 2018-10-03
Jen King - React Insta Clone

pghrogue:
@kiadorii
### Tasks (Day I)
- [x] At the end of Day I there will be a single instance of the Search Bar being rendered at the top of the page, as well as a Post Container and a Comment Section for every piece of mock data in the `dummy-data.js` file.
- [x] The root App component of your application should import the dummy data from the `dummy-data.js` file with `import dummyData from './dummy-data';` and iterate over said data, passing each individual object as a prop to an instance of `PostContainer`.
- [x] Each `PostContainer` component will then pass the array of comments on each post object as a prop to an instance of the `CommentSection` component.
- [x] The `CommentSection` component will receive the array of comments as props and render a `Comment` component with the username of the poster as well as the post's text. Additionally, there should be an input box that allows users to submit a new comment for any post. We'll work on posting new comments tomorrow.
- [x] Be sure to check the `Types` of the data you are passing around as props in the components that will be using props to present data as DOM elements. This should be linked to your `Comment` component that `Comment Section` will render and potentially to your `Post` component that `Post Container` will render.
- [x] In addition to Bootstrap for theming, you'll want to add your own styles via CSS. To keep things organized, have the CSS file that corresponds with a component live in the same directory as the component file.

### Tasks (Day II)
- [x] For this part of the project you are going to use React's lifecycle methods to get data and to render the components. Then you will use functions passed as props to build out the functionality more.
- [x] As data comes into the component, you will set it to the component's state, then pass it down to the child components.
- [x] Add the functionality to add a comment to any of the posts. Since there is no login page or anything of that sort, hard code a static username.
- [x] Implement the ability to like a post by clicking on a heart icon and having the number of likes increment accordingly.
- [x] Get the Search Bar to filter posts by the post's username. When you submit a search term should filter out posts by users whose usernames do not match the search term.

### DIFF
--- a/src/components/CommentSection/CommentSection.js
+++ b/src/components/CommentSection/CommentSection.js
@@ -1,8 +1,41 @@
 import React from 'react';
 import PropTypes from "prop-types";
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
-import './CommentSection.css';
+
 import moment from 'moment';
+import styled from 'styled-components';
+
+// Styled Components:
+const CommentSectWrapper = styled.div`
+    display: flex;
+    flex-direction: column;
+    align-items: flex-start;
+    justify-content: left;
+    padding: 10px;
+`;
+const CommentWrapper = styled.div`
+    padding-bottom: 10px;
+`;
+const DateWrapper = styled.div`
+    font-size: 11px;
+    color: gray;
+`;
+const NewComment = styled.div`
+    height: 40px;
+    margin-top: 10px;
+    border-top: 1px solid gray;
+    width: 780px;
+    display: flex;
+    align-items: center;
+    justify-content: flex-start;
+`;
+const InputWrapper = styled.input`
+    width: 730px;
+    height: 25px;
+    border: 0px solid gray;
+    margin-top: 10px;
+`;
+
 
 const CommentSection = (props) => {
     // console.log("commentsection", props);
@@ -20,22 +53,22 @@ const CommentSection = (props) => {
     let i = 0;
 
     return (
-        <div className='commentSection'>
+        <CommentSectWrapper>
             {props.comments.map( comment => {
                 return(
-                    <div className='comment' key={i++}>
+                    <CommentWrapper key={i++}>
                         <b>{comment.username}</b> {comment.text}
-                    </div>
+                    </CommentWrapper>
                 );
             }
 
             )}
-            <div className='postedAt'>{postedAt}</div>
-            <div className='newComment'>
-                <input className='commentInput' onKeyUp={props.handleComment} placeholder='Add a comment...' />
+            <DateWrapper>{postedAt}</DateWrapper>
+            <NewComment>
+                <InputWrapper onKeyUp={props.handleComment} placeholder='Add a comment...' />
                 <FontAwesomeIcon icon="ellipsis-h" />
-            </div>
-        </div>
+            </NewComment>
+        </CommentSectWrapper>
     );
 };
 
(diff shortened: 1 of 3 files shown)


pghrogue:
#### Tasks (Day III)
- [ ] Create a `<PostsPage />` component in your `components/PostsContainer` directory.
- [ ] You'll have to move a lot of what is rendered in `app.js` to this new component
- [ ] In app.js, render the `PostsPage` component.
- [ ] Make sure the app working as it was before since it has been re-factored now.
- [ ] This is to ensure that we clean up our App component a little bit before we re-factor it to be wrapped up in an HOC

- [ ] Create a directory called `Authentication`
- [ ] Inside that directory create a component called `<Authenticate />`. This is where all of the magic sis going to happen.
- [ ] This component should be able to take in a component as an argument, and it will return a `class` component.
- [ ] Inside of `<Authenticate />'s` render method, you'll want to return the Component that gets passed into it.
- [ ] Be sure to export out this component.
- [ ] Head over to App.js and `import` in our new `Authenticate` Higher Order Component and pass in `App`.
- [ ] If this worked right then everything should render as it used to.

Build out the LoginPage component. You can design it how you like
- [ ] In your `components` directory, create a directory called `Login` and add a new file called `Login.js`.
- [ ] There should be a `username` input, a `password` input, and a `Login` button.
- [ ] The component should invoke the `login` function in `app.js` when a user logs in.
- [ ] This login function should set a `username` on `localStorage`. You'll need to check local storage to see if a user is logged in.
- [ ] Be sure to force the page to reload when a user logs in so that our component un-mounts and mounts again.

Extending the functionality of the HOC to conditionally render the `LoginPage` or the `App`
- [ ] Inside of `Authenticate` we need to add a constructor to hold our state data.
- [ ] On state we need a user `loggedIn` boolean flag.
- [ ] On `componentDidMount` we need to check `localStorage` to see if a user is logged in.
- [ ] Inside of the render function we will check `if a user is logged in`
- [ ] If a user is logged in we will return the `<App />`, else we will return the `<LoginPage>`

src/components/SearchBar/SearchBar.css
```diff
@@ -0,0 +1,62 @@
...
+.logoIcon {
+    height: 29px;
+    width: 29px;
+    padding-right: 15px;
+}
+
+.logoName {
+    height: 29px;
+    align-self: flex-end;
+    border-left: 2px solid gray;
```
kiadorii: Was the border-left intentional??

src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,85 @@
...
+    // Format the timestamp using moment:
+    const postedAt = moment(props.timestamp, 'MMMM Do YYYY, h:mm:ss a').fromNow();
+
+    let i = 0;
+
+    return (
+        <CommentSectWrapper>
+            {props.comments.map( comment => {
+                return(
+                    <CommentWrapper key={i++}>
```
kiadorii: Great way of assigning a key here!

src/components/PostContainer/PostContainer.js
```diff
@@ -0,0 +1,113 @@
...
+
+const PostImgWrapper = styled.img`
+    width: 798px;
+    height: 798px;
+`;
+
+// Body of the container
+
+class PostContainer extends Component {
+    // Props taken in:
```
kiadorii: Nice visualization - I can see what you are receiving and how you are utilizing them!

PR #498, opened 2018-10-03
Kyle Teeter

kyleteeter:
Project 

Day 1 - Tasks

- [ ] At the end of Day I there will be a single instance of the Search Bar being rendered at the top of the page, as well as a Post Container and a Comment Section for every piece of mock data in the `dummy-data.js` file.
- [ ] The root App component of your application should import the dummy data from the `dummy-data.js` file with `import dummyData from './dummy-data';` and iterate over said data, passing each individual object as a prop to an instance of `PostContainer`.
- [ ] Each `PostContainer` component will then pass the array of comments on each post object as a prop to an instance of the `CommentSection` component.
- [ ] The `CommentSection` component will receive the array of comments as props and render a `Comment` component with the username of the poster as well as the post's text. Additionally, there should be an input box that allows users to submit a new comment for any post. We'll work on posting new comments tomorrow.
- [ ] Be sure to check the `Types` of the data you are passing around as props in the components that will be using props to present data as DOM elements. This should be linked to your `Comment` component that `Comment Section` will render and potentially to your `Post` component that `Post Container` will render.
- [ ] You are free to leverage the Bootstrap library for this project for the purposes of theming and styling. I recommend the awesome [reactstrap](https://reactstrap.github.io/) library, which is a library of Bootstrap components that have been implemented using React, so they're really easy to just drop straight into React projects.


eddygonzalez9708:
You do not need window.location.reload() on line 20 in your login component. It will refresh the page when you click on the log in button even when you do not submit a username or password on the login page. It also appears you have some trouble implementing a search filter. A great way to filter input values in your search handler is using an onChange handler and the search method for strings. The search method will check to see if a string exists in another string. Therefore, you can remove your event handler and filter results in your search handler as soon as the user starts typing. Just remember to display all posts when there are no characters typed in the search input field. Here is a site that contains more information about the search method:

https://www.w3schools.com/jsref/jsref_search.asp.

Also, the like button should not continue increasing the number of likes for a post. When a user clicks on the like button again, the number of likes for the post should decrease by 1. Lastly, if you want to persist user events (e.g. Liking a post and submitting a comment) you can store an object in local storage that looks similar to your local state. This object will contain all your posts, comments and likes information. As a result, a user will be able to log back in and see the same comments they submitted in the past.

Please reach out to me if you have any questions.

PR #499, opened 2018-10-03
Kendra Williams - React InstaClone

kmonique:
@kiadorii


kmonique:
Read this article about 'State Driven Views' and come up with a short paragraph as to why you feel that state-driven views are important to us as software developers today? What are state-driven applications? Why are they powerful? How does React enables those patterns?

PR #508, opened 2018-10-03
Jess Harrison React-Insta-Clone

jessharrison83:
Tasks (Day I)
-[]There are three major container components that you'll need to implement for this project: the Search Bar, the Post Container, and the Comment Section.
-[]At the end of Day I there will be a single instance of the Search Bar being rendered at the top of the page, as well as a Post Container and a Comment Section for every piece of mock data in the dummy-data.js file.
-[]The root App component of your application should import the dummy data from the dummy-data.js file with import dummyData from './dummy-data'; and iterate over said data, passing each individual object as a prop to an instance of PostContainer.
-[]Each PostContainer component will then pass the array of comments on each post object as a prop to an instance of the CommentSection component.
-[]The CommentSection component will receive the array of comments as props and render a Comment component with the username of the poster as well as the post's text. Additionally, there should be an input box that allows users to submit a new comment for any post. We'll work on posting new comments tomorrow.
-[]Be sure to check the Types of the data you are passing around as props in the components that will be using props to present data as DOM elements. This should be linked to your Comment component that Comment Section will render and potentially to your Post component that Post Container will render.
-[]You are free to leverage the Bootstrap library for this project for the purposes of theming and styling. I recommend the awesome reactstrap library, which is a library of Bootstrap components that have been implemented using React, so they're really easy to just drop straight into React projects.
-[]In addition to Bootstrap for theming, you'll want to add your own styles via CSS. To keep things organized, have the CSS file that corresponds with a component live in the same directory as the component file.


eddygonzalez9708:
Instead of using a default value for your search input field, you can use a placeholder attribute and assign it the word "search". This prevents the word "search" from being part of the input value after pressing enter. A better way to filter input values in your search handler is using the search method for strings. The search method will check to see if a string exists in another string. This will enable you to remove the event.key condition for "enter" and filter results as soon as the user starts typing. Just remember to display all posts when there are no characters typed in the search input field. Here is a site that contains more information about the search method: 

https://www.w3schools.com/jsref/jsref_search.asp. 

Also, the like button should not continue increasing the number of likes for a post. When a user clicks on the like button again, the number of likes for the post should decrease by 1. Lastly, if you want to persist user events (e.g. Liking a post and submitting a comment) you can store an object in local storage that looks similar to your local state. As a result, a user can log back in and see the same comments they type in before.

Please reach out to me if you have any questions.

PR #509, opened 2018-10-04
Bruce Cabanayan - React Insta Clone Project

bcabanayan:
This is my submission for the React Insta Clone Project


instagram/src/App.js
```diff
@@ -0,0 +1,28 @@
...
+import dummyData from './dummy-data';
+import SearchBar from './components/SearchBar/SearchBar.js';
+import PostContainer from './components/PostContainer/PostContainer.js';
+import PostsPage from './components/PostContainer/PostsPage.js';
+import Login from './components/Login/Login.js';
+import Authenticate from './components/Authentication/Authenticate.js';
+
+
+class App extends Component {
+  constructor() {
```
kiadorii: You can remove this constructor and the contents inside of it. The constructor for a React component is called before it is mounted. Since this is not called before it is mounted, there is no need for it.

instagram/src/App.js
```diff
@@ -0,0 +1,28 @@
...
+    super();
```
kiadorii: Similarly - `super()` calls the constructor of its parent class. This is required when you need to access some variables from the parent class. There is no variables that you are trying to access.

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,125 @@
...
+                {this.state.comments.map(comment => {
+                    return (
+                        <Comment 
+                            key={Math.random()} 
+                            comment={comment}
+                        />
+                    )
+                })}
+                {/* time stamp */}
+                <Timestamp>{this.props.data.timestamp.toUpperCase()}</Timestamp>
```
kiadorii: If you want to style the timestamp, I would recommend looking into regex. It will be something that you will get exposure to in the future and makes formatting simpler after understanding how it works!

instagram/src/components/PostContainer/PostContainer.js
```diff
@@ -0,0 +1,125 @@
...
+    componentDidMount() {
+        this.setState({
+            post: this.props.postInfo,
+            likes: this.props.postInfo.likes,
+        })
+    }
+
+    //determine whether there are multiples "likes" or a single "like"
+
+    checkLikes () {
```
kiadorii: I like how it is only set to increment in a plus or minus 1.

instagram/src/components/PostContainer/PostsPage.js
```diff
@@ -0,0 +1,77 @@
...
+      this.setState({
+        searchInput: e.target.value
+      })
+    }
+  
+    //perform search
+  
+    submitSearch = (e) => {
+      e.preventDefault();
+      let posts = this.state.posts.slice();
```
kiadorii: Ooh, like the usage of `slice()` here!

PR #513, opened 2018-10-04
Luis Alvarez: Instaclone

TheGameFreak720:
@juliankohlman


insta/src/App.js
```diff
@@ -0,0 +1,112 @@
...
+    });
+  }
+
+  addNewComment = (index, event) => {
+    if(event.which === 13) {
+
+      const comment = {
+        username: 'Cyanide6033',
+        text: event.target.value,
+        id: Date.now()
```
juliankohlman: Two birds with one stone here because we may want to display this data on comments, and it can serve as a key later on 👍

insta/src/components/Post.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Post = props => {
+    return (
+        <>
+            <div className='flex-column-container'>
+                <img className='thumbnail' src={props.post.thumbnailUrl} alt={'user'}></img>
```
juliankohlman: If you wanted to make your Post component stateful, what would be in the state of each post? How would you compose other components with a Post?

insta/src/components/PostContainer.js
```diff
@@ -0,0 +1,24 @@
...
+import './Posts.css';
+import PropTypes from 'prop-types';
+
+const PostContainer = props => {
+    return (
+        <div className='post-container'>
+            {props.posts.map(post => (
+                <div className='post-tunnel' key={post.timestamp}>
+                    <Post post={post} />
+                    <CommentSection comments={post.comments} likes={props.likes[post.index]} index={post.index} commentInput={props.commentInput} handleSubmit={props.handleSubmit} addNewComment={props.addNewComment} newLike={props.newLike}/>
```
juliankohlman: I'd break these props out onto separate lines as the line is running long.

PR #515, opened 2018-10-04
Jeffrey Johnson React-Insta-Clone In Progress

Gorillabuds:
@kingatoki


instagram/src/App.js
```diff
@@ -0,0 +1,28 @@
...
+import './App.css';
+import dummyData from './dummy-data';
+
+class App extends Component {
+  render() {
+    return (
+      <div>
+        <h1>instagram</h1>
+        <span>
+          {dummyData.map((post,i)=><div key={i}><PostContainer container={post}/></div>)}
```
KingAtoki: I would avoid using the index as a key. I would prop doing something like`key={post + i}`

instagram/src/App.js
```diff
@@ -0,0 +1,28 @@
...
+        </span>
+        <div><PostContainer/></div>
```
KingAtoki: Since you are mapping over dummy data and filling in your posts that way, you probably don't need this, as it's going to be blank

instagram/src/components/PostContainer.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+
+class PostContainer extends React.Component{
+    render(props){
```
KingAtoki: No need to pass anything to the render method

instagram/src/components/PostContainer.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+
+class PostContainer extends React.Component{
```
KingAtoki: This can be a functional component since you aren't using state or any of the lifecycle methods

PR #518, opened 2018-10-04
Mitchell Robles - React-Insta-Clone

mitchellr92:
@John-Spraul


instagram/src/components/SearchBar/SearchLogo.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import './SearchBar.css';
+
+const SearchLogo = (props) => {
+    return (
+        <div className='search-logo'>
+            <i className="fab fa-instagram fa-2x"></i>
+            <div className='divider'> | </div>
```
John-Spraul: did you consider adding a border instead of using `|` ?

PR #524, opened 2018-10-04
Angelina La Salle React-Insta-Clone

alasalle:
@juliankohlman


instagram/src/components/Comment/Comment.js
```diff
@@ -0,0 +1,32 @@
+import React from "react";
+import PropTypes from "prop-types";
+import '../Comment/Comment.css';
+
+class Comment extends React.Component {
+  render() {
+    return (
+      <div className={'commentItem'}>
```
juliankohlman: Your `classNames` don't need to wrapped with curly braces. Anything wrapped in curly braces in JSX will be evaluated as a JS expression. If you were generating classNames dynamically that'd be different, but in that case, you'd be passing a variable without quotes into your curly braces.

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,111 @@
...
+          <img
+            src="https://png.icons8.com/ios/48/000000/bookmark-ribbon.png"
+            alt=""
+          />
+        </div>
+        <div className={"likesCount"}>{this.state.likes} likes</div>
+        {this.state.comments.map(comment => {
+          return (
+            <Comment
+              key={Math.random()}
```
juliankohlman: You could access each comments index from the comments array instead of using Math.random to generate a key. Another idea would be to slice off n amount of letters from a comment as a key, but here's some more info about Keys straight from Reacts documentation:

`The best way to pick a key is to use a string that uniquely identifies a list item among its siblings. Most often you would use IDs from your data as keys:`
```
const todoItems = todos.map((todo) =>
  <li key={todo.id}>
    {todo.text}
  </li>
);

```
`When you don’t have stable IDs for rendered items, you may use the item index as a key as a last resort:`

instagram/src/components/Post/Post.js
```diff
@@ -0,0 +1,46 @@
...
+import PropTypes from "prop-types";
+import '../Post/Post.css';
+
+class Post extends React.Component {
+  render() {
+    return (
+      <div className={"postContainer"}>
+        <div className={"postTop"}>
+          <img src={this.props.post.thumbnailUrl} alt="user thumbnail" />
+          <div className={'topUsername'}>{this.props.post.username}</div>
```
juliankohlman: Unless we need JSX to evaluate some JavaScript for us, we don't need curly braces for our classNames. Also be careful about mixing double and single quotes, it's best to pick one and stick with it.

PR #529, opened 2018-10-05
Insta-Clone

ShardeDeLane:


instagram/src/Components/Authentication/Authenticate.js
```diff
@@ -0,0 +1,35 @@
...
+			} else {
+				this.setState({
+					loggedIn: false
+				})
+			}
+		}
+		render() {
+			if (this.state.loggedIn === true) {
+				return <App />
+			} else if (this.state.loggedIn === false) {
```
Raymondgrc: Another syntax to this statement is removing the else if, and just adding `return <Login /> ` immediately after the end brace of the if statement.

instagram/src/Components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,44 @@
...
+	constructor(props) {
+		super(props)
+		this.state = {
+			comments: props.comments
+		}
+	}
+
+	addComment = newcomment => {
+		const commentAdded = this.state.comments
+		localStorage.setItem('username', 'rhcprocker')
```
Raymondgrc: Later, I would recommend grabbing the username from the localstorage from the login page

instagram/src/Components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,44 @@
...
+			comments: commentAdded
+		})
+	}
+
+	render() {
+		return (
+			<CommentSectionWrapper>
+				<Likes likes={this.props.likes} />
+				{this.state.comments.map((comment,index) => (
+					<Comments key={index} comment={comment} />
```
Raymondgrc: Remember to avoid setting an index as the key, in case the order changes later.

PR #535, opened 2018-10-06
Zachary Arney

zacharyarney:


instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,62 @@
...
+
+  addNewComment = (event) => {
+    event.preventDefault();
+    console.log('before function', this.state.comments);
+    if (this.state.newComment.length > 0) {
+      this.setState({
+        comments: [
+          ...this.state.comments,
+          {
+            username: localStorage.getItem('username'),
```
luiscmartinez: good thinking how to use the local storage data to your advantage

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,62 @@
...
+
+  changeHandler = (event) => {
+    this.setState({ [event.target.name]: event.target.value });
+  };
+
+  render() {
+    return (
+      <CommentSectionContainer>
+        {this.state.comments.map((comment) => {
+          return <Comment comment={comment} key={comment.text} />;
```
luiscmartinez: I still give you credit for not using only the index as key
setting our key to the index will forever be our last resort but we shall probably never do it
We are save to assume that users usernames will all be unique

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,62 @@
...
+        })}
+        <div className="timestamp">{this.props.timestamp}</div>
+        <form autoComplete="off" onSubmit={this.addNewComment}>
```
luiscmartinez: cheeky autoComplete attribute
wasn't aware of such thing, thanks!

instagram/src/components/PostContainer/PostsPage.js
```diff
@@ -0,0 +1,78 @@
...
+        (post) => post.username === this.state.searchQuery
+      ),
+    });
+  };
+
+  addLike = (timestamp) => {
+    this.setState({
+      posts: this.state.posts.map((post) => {
+        if (post.timestamp === timestamp) {
+          post.likes += 1;
```
luiscmartinez: I can like a thousand times . 
my challenge for you is to implement a way where the user is only allowed to like a post once

PR #539, opened 2018-10-08
Matt Wright FSW14 React-Insta-Clone

mattwright42:
initial pull request


instagram/src/components/Authentication/Authenticate.js
```diff
@@ -0,0 +1,27 @@
...
+    componentDidMount() {
+      if (localStorage.getItem('user')) {
+        this.setState({ isLoggedIn: false });
+      } else {
+        this.setState({ isLoggedIn: true });
+      }
+    }
+
+    render() {
+      if (this.state.isLoggedIn) return <Login />;
```
cesarnml: the logic seems a little reversed ... it looks like if `isLoggedIn is true ... you show the <Login/> component but logically you would want <Login/> to be returned if isLoggedIn is false. It works ... just reads funny.

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,72 @@
...
+
+import './CommentSection.css';
+
+class CommentSection extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      comments: props.comments,
+      newComment: '',
+      likes: 0
```
cesarnml: It would be nice if likes was dynamically set to the starting value of dummyData

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,72 @@
...
+    });
+  };
+
+  handleChange = event => {
+    this.setState({ [event.target.name]: event.target.value });
+  };
+
+  incrementLikes = event => {
+    event.preventDefault();
+    this.setState({ likes: this.state.likes + 1 });
```
cesarnml: it seems odd that the likes just increment without end. typically you would be allowed to like/unlike on each click.

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,72 @@
...
+            addNewComment={this.addNewComment}
+            handleChange={this.handleChange}
+          />
+        </div>
+      </div>
+    );
+  }
+}
+
+CommentSection.propTypes = {
```
cesarnml: nice consistent prop typechecking

instagram/src/components/Login/Login.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  handleInputChange = event => {
+    this.setState({ [event.target.name]: event.target.value });
+  };
+
+  loginAttempt = event => {
+    const user = this.state.username;
+    localStorage.setItem('user', user);
+    window.location.reload();
```
cesarnml: nice job managing to persist login on refresh

instagram/src/App.js
```diff
@@ -0,0 +1,25 @@
...
+import PostsPage from './components/PostContainer/PostsPage';
+import Authenticate from './components/Authentication/Authenticate';
+
+import './App.css';
+
+class App extends Component {
+  constructor() {
+    super();
+    this.state = {
+      posts: [],
```
cesarnml: i don't believe posts on app is being used now that you have PostsPage or search.

instagram/src/components/PostContainer/PostsPage.js
```diff
@@ -0,0 +1,50 @@
...
+    };
+  }
+
+  componentDidMount() {
+    this.setState({ posts: dummyData });
+  }
+
+  search = event => {
+    const newPosts = this.state.posts.filter(post => {
+      if (post.username.includes(event.target.value)) {
```
cesarnml: I believe this can be reduced to `return post.username.includes(event.target.value)` since filter just needs to return true or false to decide whether or not to return post and `.includes` by default returns a boolean.

PR #545, opened 2018-10-08
Conner H [React-insta-clone]

TRIF3X:


CookieMonsta89:
Hey Conner, you are doing good so far.  Some advice that I would give is that you have too many specific things being passed down from your app.js state.  Consider passing all of the dummydata to your components then breaking it up as you pass them around to other components.  It isn't a bad thing that every component has access to the props even if they don't use it.  I would also advise against having so many functions on your app.js.  I was once told that you want the app.js to be as clean as possible.  You do this by breaking components down into smaller components and passing data rather than having everything on the app.js.  For example, Post container could probably be broken into 3-4 more components such as PostHeader, Posts, and Likes Section.  Keep working hard.

CookieMonsta89:
Hey Conner, so the app looks great but there are a few issues going on here.  First off, your search bar placeholder says username.  Second, I'm not sure if the addnewcomment function is being passed correctly because when you go to add a new comment to a card, it reloads the page which prevent.default should help with.  You are getting really close so keep pushing.

PR #548, opened 2018-10-08
Peter Pham - React-Insta-Clone

phamhere:


instaclone/src/components/Login/Login.js
```diff
@@ -0,0 +1,57 @@
...
+            name="password"
+            placeholder="Password"
+            onChange={this.handleInputChange}
+            value={this.state.password}
+            required
+          />
+          <input
+            type="submit"
+            value="Sign up"
+            onClick={this.handleLoginSubmit}
```
luiscmartinez: you should give the submit method to the Form tag as an onSubmit => therefore the submit button will be bound to the submit when click on this input or when the user presses enter. 
Currently when I try to submit with my enter key => the users loses their credentials data & nothing happens

instaclone/src/App.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import './App.css';
+import Authenticate from './components/Authentication/Authenticate';
+import PostsPage from './components/PostContainer/PostsPage'
+
+class App extends React.Component {
+  constructor() {
+    super();
+    this.state = {
+      username: ''
```
luiscmartinez: why do we need this on state?`

instaclone/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,74 @@
...
+      ];
+      if (e.key === "Enter") {
+        this.setState({
+          comments: newComments,
+          comment: ""
+        });
+      }
+    }
+  };
+  deleteComment = index => {
```
luiscmartinez: good job implementing this
my challenge to you is to change this methods functionality so only the comments that belong to the user can be deleted

instaclone/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,74 @@
...
+    let filteredComments = this.state.comments.filter(
+      (comment, i) => index !== i
+    );
+    this.setState({ comments: filteredComments });
+  };
+  onCommentChange = e => {
+    this.setState({ [e.target.name]: e.target.value });
```
luiscmartinez: an alternative using destructuring could be ==>

 ({ target }) => {
      const { name, value } = target
      this.setState({[name]: value})
    }

instaclone/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,74 @@
...
+  componentDidUpdate() {
+    localStorage.setItem(`post${this.state.index}`, JSON.stringify(this.state));
+  }
+  render() {
+    return (
+      <React.Fragment>
+        {this.state.comments.map((comment, i) => (
+          <Comment
+            comment={comment}
+            key={i}
```
luiscmartinez: using the index as key should always be our last resort . 
We should always try our best to avoid using only the index by itself as the key .

We are save to assume that a user's username will be unique throughout the website

instaclone/src/components/PostContainer/PostContainer.js
```diff
@@ -0,0 +1,70 @@
...
+        : this.props.data.likes,
+      postLiked: localStorage.getItem("liked")
+        ? JSON.parse(localStorage.getItem("liked"))
+        : false
+    };
+  }
+  plusLikes = () => {
+    this.setState(prevState => {
+      return {
+        likes: prevState.postLiked ? prevState.likes - 1 : prevState.likes + 1,
```
luiscmartinez: hack proof !

instaclone/src/components/PostContainer/PostContainer.js
```diff
@@ -0,0 +1,70 @@
...
+        postLiked: !prevState.postLiked
+      };
+    });
+  };
+  componentDidUpdate() {
+    localStorage.setItem("likes", this.state.likes);
```
luiscmartinez: good stuff persisting the likes too !

PR #550, opened 2018-10-09
React-Insta-Clone (Sarah Tennis)

sarahtennis:


Ta1grr (commented):
## Great
 - Unique take on the login screen.
 - Really like how when you type in for comments the text are gray
## Requested Improvements
 - Add in comments to explain your workflow.
## Questions
 - I remember during the evening stand-up regarding reusability with styled-components, It's very similar to using Parent Selector in LESS. It'll use the same button, but depending on what you pass as props in this case will determine the output of the HTML element.
 - Also, not much to comment on this project, it is working as intended.
## Rating: {1-3}
 - 2 - Understands with assistance, meets expectations

PR #552, opened 2018-10-09
Hargobind Atwal

ethiapath:


CookieMonsta89:
Hey Hargobind, I took a look at your project today and it's possible you have not pushed anything new to the project because it is on your machine.  I can't really add any new advice but I can say that your structure of files looks really great.  I like that you are keeping everything broken up very well.  I can see you are testing the add comment function so I'm sure that will work very soon.  Just keep working on it because you are doing great.

PR #553, opened 2018-10-09
Jordan Stoddard - React-Insta-Clone Project

Jordan-Stoddard:


instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,85 @@
...
+    
+
+      changeHandler = event => {
+        this.setState({[event.target.name]: event.target.value });
+      }
+
+      addNewComment = (event, index) => {
+          
+        event.preventDefault();
+        if (this.state.commentInput === '') {
```
cesarnml: `.trim()` might accomplish the same effect but prevent also blank inputs. The user at this point is able to add a comment that is blank if he hits spacebar

PR #555, opened 2018-10-09
Ryan Clausen

theryanclausen:


CookieMonsta89:
Hey Ryan, you've really impressed me with your work on this project so far.  You are definitely doing some really great things here.  I love that you created instances of the images you needed and passed them through as instances rather than objects.  This was a really cool idea.  You app so far is not glitchy and works seamlessly.  For someone to not have worked React before, you're doing great.  It is kind of difficult to find something to critique but if there is any advice I can give you it is definitely to keep stretching on these assignments.  I firmly believe you do your best learning when trying to stretch and it forces you to proactively research.  So keep up the good work.  I also like that the like button doesn't increment everytime and only does the once and can reset the state back to the original amount.  Maybe adding a double click on the image for a like would be a great stretch.  Or even having the search bar work for comments and users as well.

CookieMonsta89:
Hey Ryan, this is awesome.  I like that you added a new card, what a great stretch.  Everything is very seamless and the login page's design is very cohesive with the instagram page.  You handled the authentication process very clean and dry so great job on that.  I like that you can click on a post to bookmark it.  It would be a cool idea if when you selected the heart at the top of the page, that it only displayed your bookmarked cards.  Just another stretch since you're knocking them out of the park.  Keep up the good work.

PR #557, opened 2018-10-09
Yanrong_Xu - React-Insta-Clone

YanrongXu:
@kamry-bowman


instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,66 @@
...
+import PropTypes from "prop-types";
+import AddComment from "./AddComment";
+import "./CommentSection.css";
+
+class CommentSection extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      comments: props.user.comments,
+      timestamp: props.user.timestamp,
```
kamry-bowman: not clear why this is on `state`, since it doesn't change. Only put properties that would change on `state`

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,66 @@
...
+      ];
+
+      this.setState({
+        comments: newCommentList,
+        newComment: ""
+      });
+    }
+  };
+
+  changeHandler = (key, value) => {
```
kamry-bowman: nice job making this general!

instagram/src/components/PostContainer/PostsPage.js
```diff
@@ -0,0 +1,73 @@
...
+    this.state = {
+      postData: [],
+      newSearch: "",
+      filtered: false
+    };
+  }
+   componentDidMount() {
+    setTimeout(() => {
+      this.setState({ postData: dummyData });
+    }, 500);
```
kamry-bowman: great job making this async

instagram/src/components/PostContainer/PostsPage.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+   startSearch = event => {
+    event.preventDefault();
+    let newData = dummyData;
```
kamry-bowman: newData should be taken off state, not the original dummyData. What if a post got deleted for example? The post would keep coming back in your filtered data this way, which isn't good.

instagram/src/components/PostContainer/PostsPage.js
```diff
@@ -0,0 +1,73 @@
+import React from "react";
+import dummyData from "../../dummy-data";
+import PostContainer from "./PostContainer";
+import SearchBar from "../SearchBar/SearchBar";
+ class PostsPage extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      postData: [],
```
kamry-bowman: you would normally want to have a separate postData on state to keep track of all of your posts. The way you have this set up is going back to the dummyData each time you filter, but you would want to have your postData be dynamic from dummyData. A good solution would be to use a separate filteredData in addition to the postData item, and show the filteredData list

PR #560, opened 2018-10-09
React-Insta-Clone By: Yusuf Nafey

yusufnafey:


KevinBrack:
The project is off to a great start. Lots of functionality present: Likes, Search, icons. Doing a great job. Keep it up. Not seeing any styling at this point, but function over form seems to be working for you.

KevinBrack:
Site looks and functions great. Nice work with the login HOC. I would caution against using the class name "App" outside of the app.js (found in PostsPage component) because that is what React is using as a mounting point for the application. Nice work.

PR #562, opened 2018-10-09
React-Insta-Clone -- Brian Williams

bkwillia052:
@kamry-bowman


instagram/src/App.js
```diff
@@ -0,0 +1,109 @@
...
+library.add(faUserCircle);
+
+
+class App extends Component {
+  constructor(props){
+    super(props);
+    this.state = {
+      stateComments: [],
+      commentInput: {},
+      dummyData: [],
```
kamry-bowman: don't name it dummyData on state... that's the source of the data, but it is the actual application's data with its on state. Name it posts or something...

instagram/src/components/Authentication/Authenticate.js
```diff
@@ -0,0 +1,71 @@
...
+        } 
+        
+            
+        
+    }  
+
+    render(){
+        return (
+        <div>
+        {window.localStorage.getItem('loggedIn') ? <App username={this.state.username} logOut={this.logOut} /> : <Login login={this.login} onKeyDown={this.inputHandler}/>}
```
kamry-bowman: use this.state for your conditional, that way when you update state to reflect being logged in, you will automatically switch to rendering App.

Also, be sure to spread props onto App. Like this: `<App {...props} />`

instagram/src/App.js
```diff
@@ -0,0 +1,109 @@
...
+
+library.add(faHeart);
+library.add(faUserCircle);
+
+
+class App extends Component {
+  constructor(props){
+    super(props);
+    this.state = {
+      stateComments: [],
```
kamry-bowman: seems like stateComments, commentInput aren't getting used, so you can probably take them off state

PR #564, opened 2018-10-09
Melinda Yeager

Mindysueasaurus:


CookieMonsta89:
Great job Mindy on passing your props and getting them to render.  I feel that you have done a great job also at structuring your files efficiently.  I like when components are broken down into even smaller components.  It makes things easy to follow and shows that you understand how the passing of data works.

CookieMonsta89:
Hey Mindy, took a look at your project and a few things are breaking it.  I would create an Authenticate.js file and have that deciding whether or not it should display the app or the login page depending on if someone if logged in or not.  Right now, you are importing authenticate but I don't see where there is an authenticate file.

PR #565, opened 2018-10-09
Rahul Desai [React Insta Clone]

rd825:


instagram/src/components/Login/Login.js
```diff
@@ -0,0 +1,81 @@
...
+`;
+
+const StyledHeader = styled.h2`
+    color: lightgrey;
+    font-weight: bold;
+    font-size: 14px;
+    padding: 0px 40px;
+`;
+
+
```
Ta1grr: Although normal circumstances you'll test to see if the user exist with a database, but you could do a set-up a similar way with a localstorage.

instagram/src/components/PostContainer/PostsPage.js
```diff
@@ -0,0 +1,39 @@
...
+class PostsPage extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      postData: dummyData,
+      searchString: '',
+    }
+  }
+
+  searchHandler = event => {
```
Ta1grr: For some reason, it won't pull up fortnite in the search, it works for the other two cards though. I even test it on your netlify as well.

rd825: Yeah my search was behaving badly but I couldn't figure out what was wrong or how to resolve...

Ta1grr: So after looking over your code this morning, I finally got the `<Searchbar/>` to update state in your `<Postpage/>`, but the problem I'm having is you having the logic set-up within `<Post/>`. That kinda defeat the purpose of reusability because that component can only fulfill one function, a solution would be creating a handler inside your state component which would be `<Postpage/>`. So that way you can `.filter` through things and adjust state accordingly save it somewhere so the page can render the result from `.filter` method.

rd825: So the way filter was done on the Sprint Challenge today has opened my eyes to what I should have been doing.

PR #569, opened 2018-10-09
Sean Won

swon8:


CookieMonsta89:
Hey Sean, so my advice so far would be to just jump right in on the project and do what you do know.  It is easy to get stuck in this cycle of researching that you can fall behind because you are worried that you don't know enough to start.  The best way to learn is by making mistakes and doing.

CookieMonsta89:
Hey Sean, good job jumping straight in and getting started.  Looks like you have everything displayed correctly so you are mapping correctly which is great.  It is kind of okay that you don't have styling done as you will get to start styling with styled components after lecture today.

PR #570, opened 2018-10-09
Instaclone Project

sabrinagear:


KevinBrack:
Nice work. Styling turned out great. Still not seeing any sign of those icons you worked so hard on. Nice work with styled-components and using named exports/imports. Great HOC for the login. Keep this up!

PR #574, opened 2018-10-09
Kyran McCann

kyranmccann:
Day 2 - functionality is there, still need to work on styling and cleaning up some code.


instagram/src/components/PostContainer/PostsPage.js
```diff
@@ -0,0 +1,35 @@
...
+      search: '',
+    }
+  }
+
+  componentDidMount() {
+    this.setState({ dummyData: dummyData});
+  }
+
+  handleSearch = event => {
+    this.setState({[event.target.name]: event.target.value})
```
luiscmartinez: an alternative could be to destructor the parameter & the target. 
call the parameter target so the we make use of the es6 object shorthand notation 
({ target }) => {
      const { name, value } = target
      this.setState({
        [name]: value
      })
    }

PR #579, opened 2018-10-09
Elvis Ibarra - React-Insta-Clone

ielvisd:
@Sumi419, 👋


instagram/src/components/PostsContainer/LikeSection.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+
+const LikeSection = props => {
+  return [
+    <div
+      className="like-section"
+      key="likes-icons-container"
+      onClick={props.incrementLike}
```
Velsu: I would probably move it to only work when `heart` is clicked, instead of the whole div object.

PR #586, opened 2018-10-09
Instaclone - Edd Burke

bummings:


KevinBrack:
Site looks okay. Think there is a little work yet to be done on styling. Images and paragraph seem to be different widths. Nice use of prop-types and I see a little lifecycle method experimentation as well. Over all nice work.

PR #596, opened 2018-10-10
RileyB-FSW14-Instagram-React-Clone

Riley-Brown:


instagram/src/App.js
```diff
@@ -0,0 +1,22 @@
...
+import "./App.css";
+// import SearchBar from "./components/SearchBar/SearchBar";
+// import PostsPage from "./components/PostContainer/PostsPage";
+import Auth from "./components/Authentication/Authenticate";
+
+class App extends Component {
+  constructor() {
+    super();
+    this.state = {
+      loggedIn: false
```
luiscmartinez: does this ever become True ?

instagram/src/components/Authentication/Authenticate.js
```diff
@@ -0,0 +1,83 @@
...
+import PostsPage from "../PostContainer/PostsPage";
+
+class Auth extends React.Component {
+  constructor() {
+    super();
+    this.state = {
+      loggedIn: false,
+      hello: "Hello",
+      filteredPosts: [],
+      filterTarget: "",
```
luiscmartinez: its not wrong but for better name convention, name it what it is. 
An ideal name would be searchInput since that is what it is. So when other ppl read your code it is more straightforward to what its for & whats it doing .

instagram/src/components/Authentication/Authenticate.js
```diff
@@ -0,0 +1,83 @@
...
+      loginTarget: ""
```
luiscmartinez: Same with this input
Ideal name ==> usernameInput ?

instagram/src/components/CommentSection/Comment.js
```diff
@@ -0,0 +1,91 @@
...
+  componentDidMount() {
+    this.setState({
+      comments: this.props.comments,
+      likes: this.props.likes
+    });
+  }
+
+  handleLikes = e => {
+    this.setState(prevState => ({
+      likes: prevState.likes + 1
```
luiscmartinez: I can like uh million times, hack ?

luiscmartinez (commented):
Good stuff Riley, using styled components & styling this whole application. Login & Comment functionally feels & looks good. 
I want to give you the challenge of somehow stopping the user from liking more than once.

instagram/src/App.js
Riley-Brown: When there is a user in local storage it becomes true 
`if (localStorage.getItem("user")) {
      this.setState({
        loggedIn: true
      });
    } else {
      this.setState({
        loggedIn: false
      });
    }`

PR #598, opened 2018-10-10
React-Insta-Clone- Brian Gregorius

brianmgre:
@CookieMonsta89


CookieMonsta89:
Awesome job on this.  I really like the design of the landing login page. I like that it searches in real time depending on what is being typed rather than typing and hitting enter.  Good job with passing props through authenticate to render app of login. I know this week has been tough but great work Brian on pushing through.

PR #600, opened 2018-10-10
React-Insta-Clone by Katy Hollobaugh

KHollobaugh:


KevinBrack:
Having some problems viewing this site at the moment. I think maybe you are installing dependencies globally again. React-strap was properly added to the package.json but bootstrap and styled-components are still not in the dependency list. Use `yarn add bootstrap styled-components` to resolve that in the future. Also some of the file paths are broken related to css. Comment out the unused imports if you changed to styled-components. Keep up the great work.

PR #609, opened 2018-10-11
CJ Wright

cwright0428:
New PR


instagram/src/Components/PostContainer/PostPage.js
```diff
@@ -0,0 +1,49 @@
...
+class PostPage extends React.Component {
+  constructor() {
+    super();
+    this.state = {
+      postList: [],
+      commentField: ""
+    };
+  }
+  componentDidMount() {
+    this.setState({ postList: dummyData });
```
cesarnml: this could be done inside of constructor.

instagram/src/App.js
```diff
@@ -0,0 +1,49 @@
...
+    this.setState({ credentials });
+  };
+
+  handleChange = event => {
+    this.setState({ [event.target.name]: event.target.value });
+  };
+
+  handleSubmit = event => {
+    event.preventDefault();
+    const authKey = `${this.state.username} is logged in`;
```
cesarnml: it might be more useful to have `authKey = this.state.username` that way you can use it to dynamically set the username when adding comments.

instagram/src/App.js
```diff
@@ -0,0 +1,49 @@
...
+    this.setState({ username: "", password: "" });
+  };
+
+  render() {
+    return (
+      <div className="App">
+        <SearchBar handleCreds={this.handleCreds} />
+        <PostPageAuthed
+          credentials={this.state.credentials}
+          username={this.state.username}
```
cesarnml: I think username isn't being set correctly. From reading the code i bet it's always an empty string when passed down. same for password.

instagram/src/Components/Authentication/Authenticate.1.js
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import Login from "../Login/Login";
+
+export default PassedComponent => {
+  return class AuthenticationClass extends React.Component {
+    componentDidUpdate() {
+      if (!this.props.credentials && localStorage.getItem("credentials")) {
```
cesarnml: isn't this logic redundant?  Because this.props.credentials was set to localStorage.getItem('credentials') in the parent component so the left and right of the `&&` are always the same value. Not sure about this.

instagram/src/Components/Authentication/Authenticate.js
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import Login from "../Login/Login";
+
+export default PassedComponent => {
+  return class AuthenticationClass extends React.Component {
+    componentDidUpdate() {
+      if (!this.props.credentials && localStorage.getItem("credentials")) {
```
cesarnml: why two auths? same issue as auth1

instagram/src/Components/PostContainer/PostPage.js
```diff
@@ -0,0 +1,49 @@
...
+  updateCommentField = event => {
+    this.setState({ [event.target.name]: event.target.value });
+  };
+
+  submitNewComment = (event, timestamp) => {
+    event.preventDefault();
+    const newPostList = this.state.postList.map(eachPost => {
+      if (eachPost.timestamp === timestamp) {
+        eachPost.comments.push({
+          username: "Clark Kent",
```
cesarnml: i think you should try replacing the hardcoded `username: 'Clark Kent' with  the username on localStorage

instagram/src/Components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,39 @@
...
+        <div className="icons">
+          <div className="social-wrapper">
+            <i className="far fa-compass" />
+            <i className="far fa-heart" />
+            <i className="far fa-user-circle" />
+          </div>
+        </div>
+      </SearchBarWrapper>
+
+      <button onClick={() => props.handleCreds(null)}>LOGOUT!!!</button>
```
cesarnml: good job on incorporatating a log out ... you could have this trigger on `user-circle` to mimic closer the real instagram

PR #612, opened 2018-10-12
Katherine Rust Insta-Clone

fbt328:
first PR

### DIFF
--- a/DAY_TWO_README.md
+++ b/DAY_TWO_README.md
@@ -1,6 +1,6 @@
 ## Day II
 
-### Focus (Day II)
+### x Focus (Day II)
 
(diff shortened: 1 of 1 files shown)


fbt328:
@eddygonzalez9708

eddygonzalez9708:
Please reach out to me if you are having trouble getting the ball rolling with this sprint. Wrapping your imported components in div tags to style them in your App.js file is not necessary. You can style JSX code in a component using its own CSS file (e.g. SearchBar.css). Also, each component file should be in its separated directory within the components folder (e.g. SearchBar.js and SearchBar.css should be in /components/SearchBar/). The image tags in the SearchBar.js file should not have a closing tag because they are empty tags and are self-closing. Lastly, always look in your console for errors. You currently have a few error messages in your console.

PR #619, opened 2018-10-15
Joe Bugajski | React-Insta-Clone

JoeBugajski:


Wdifulvio523:
Very good start on the project so far Joe!  You have your app beginning to be wired up, and you are displaying your components. Now we need to grab the data and pass it to those components, as needed. Keep it up!

PR #620, opened 2018-10-15
M. Malik

mm2079:


Wdifulvio523:
Project is looking good so far! successfully passed down the data as a prop and mapped through it, well done. Now we just have to keep building out the components and passing the data, but you're in a very good spot so far. Keep it up!

react-insta-clone/src/App.js
```diff
@@ -12,28 +12,24 @@ class App extends Component {
...
   {
-    setTimeout( () =>
+    let someVar = document.querySelector( ".App" );
+    someVar.addEventListener( "click", ( event ) =>
     {
-      this.setState({
-        loaded: true
-      })
-    }, 1 )
+      console.log( "Hello" );
```
Wdifulvio523: looks good man, well done!

PR #621, opened 2018-10-15
Govi Escudero / React-Insta-Clone

jaklundun:


Wdifulvio523:
good start to the project! Now let's build out the components and figure out how to pass the data down as a prop to those components. Keep it up!

PR #622, opened 2018-10-15
Nick Kruger - React-Insta-Clone

NickKruger:


Wdifulvio523:
Very good start so far! you're close to completing the mapping of the data. Remember to use your react devtools to make sure the data is being passed down correctly!

PR #623, opened 2018-10-15
React-Insta-Clone - Scott Bren

smbren:
React-Insta-Clone - Scott Bren


Wdifulvio523:
you're off to a good start here, jut keep wire-framing and building out your componenets. Let me know if you need anything!

PR #640, opened 2018-11-13
Latifah President 

latifahpresident:


Synch3:
Latifah, in our zoom we talked about creating the search functionality. One thing I want to you to try and focus on is remembering when we are accessing state. Every time you want to think about modifying a component, consider whether state will be useful! We store out state on the state object to make react apps dynamic.

PR #642, opened 2018-11-13
First Submission - MVP Completed

gittc100:


instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+import './CommentSection.css';
+import PropTypes from 'prop-types';
+
+const CommentSection = props => {
+    return(
+        <div className="comment-container">
+            <p><span>{props.data.username}</span> {props.data.text}</p>
```
johnoro: Imagine how much cleaner this could be with [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring):
```javascript
const CommentSection = ({ data }) => {
  const { username, text } = data;
  return (
    <div className="comment-container">
      <p>
        <span>{username}</span>
        {text}
      </p>
    </div>
  );
};
```

instagram/src/components/PostContainer/PostContainer.js
```diff
@@ -0,0 +1,22 @@
...
+    <div className="post-container">
+      {props.data.map(item => {
+        return <PostContent key={item.username} data={item} />;
+      })}
+    </div>
+  );
+};
+
+PostContainer.propTypes = {
+    data: PropTypes.array.isRequired
```
johnoro: Here, instead of just `PropTypes.array`, you could check for a the contents of the array, to make sure each item is an object and has a username property (and whatever else it's supposed to have).

instagram/src/components/PostContainer/PostContainer.js
```diff
@@ -1,22 +1,58 @@
...
+
+    {/* <div className="post-container">
+      {props.data.map(item => {
+        return <PostContent 
+        key={item.username} 
+        data={item}
+        comments={item.comments}
+        likes={item.likes}   
+        />;
+      })}
```
johnoro: Take out the old code!

instagram/src/components/PostContent/PostContent.js
```diff
@@ -1,63 +1,115 @@
...
                 </div>
-                <div className="add-comment">
-                    <input type="text" className="input-comment" placeholder="Add a comment..."/>
+                <form  className="add-comment" onSubmit={props.addItem}>
+                    <input 
+                    className="input-comment" 
+                    placeholder="Add a comment..."
+                    type="text" 
+                    name="inputText"
+                    value={props.inputText}
```
johnoro: Remove your old code!

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -1,21 +1,131 @@
...
+        comments: newComment,
+        inputText: ""
+      });
+    }
+  };
+
+  render() {
+    // console.log(this.state.comments);
+    // console.log(this.state.inputText);
+    // console.log(this.state.index);
```
johnoro: Remove these commented out logs!

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -1,21 +1,131 @@
...
+  componentDidMount() {
+    const comments = JSON.parse(localStorage.getItem("comments"));
+    if (comments) {
+      this.setState({ comments: comments });
+    }
+  }
+
+  handleChange = event => {
+    this.setState({
+      [event.target.name]: event.target.value
```
johnoro: You could use [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring) here; for example:
```javascript
handleChange = ({ target }) => {
  const { name, value } = target;
  this.setState({ [name]: value });
};
```

instagram/src/components/Authentication/Authenticate.js
```diff
@@ -0,0 +1,67 @@
+import React from "react";
+import Login from "../Login/Login";
+
+const authenticate = WrappedComponent => {
```
johnoro: Since this is still _technically_ a component, you should still capitalize it (`Authenticate`)

PR #648, opened 2018-11-13
Eileen Cuevas

eileencuevas:


instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,83 @@
...
+    font-size: 1.1rem;
+`
+
+
+
+class CommentSection extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            comments: this.props.comments,
```
trevorfehrman: This is probably an anti-React pattern.  Bolting things onto a child component state object from it's props object can cause issues should probably be avoided.  In addition, in this case you have the data you need on the props object already making this step unnecessary.

instagram/src/index.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import './index.css';
+import App from './App';
+import * as serviceWorker from './serviceWorker';
```
trevorfehrman: If you're not using this it's a better practice to take it out, it's leaving an error in the console.

trevorfehrman: Delete the file as well

instagram/src/App.js
```diff
@@ -0,0 +1,69 @@
...
+  constructor() {
+    super();
+    this.state = {
+      data: [],
+      searchText: '',
+      isSelected: false,
+    }
+  }
+
+  logOut = () => {
```
trevorfehrman: This is a really cool detail gj.

instagram/src/App.js
```diff
@@ -0,0 +1,69 @@
...
+    localStorage.removeItem('username');
+    window.location.reload();
+  }
+
+  clearSearchText = () => {
```
trevorfehrman: You don't appear to be using this anywhere but you shouldn't need a separate method for this anyway.  On a setState in which you're using the input value you can set the string to empty as an additional expression.

instagram/src/components/PostContainer/PostContainer.js
```diff
@@ -0,0 +1,57 @@
...
+
+    img {
+        margin-top: 2px;
+    }
+`
+
+const PostContainer = props => {
+    return(
+        <PostContainerDiv>
+            {props.data.map((currentPost, index) => (
```
trevorfehrman: If you use filter here instead of map you wouldn't need that searchData method, you could simply use the search term inside your truth check here to accomplish your search functionality.

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,83 @@
...
+            timestamp: this.props.postTimestamp,
+            newComment: '',
+        }
+    }
+
+    addNewComment = event => {
```
trevorfehrman: Writing functionality this low into the component tree is questionable as it makes the application more brittle by violating the single source of truth.  I would try to lift this state to App.js and control the comment sections from there

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,83 @@
...
+
+    render() {
+        return(
+            <div>
+                {this.state.comments.map((item, index) => (
+                        <IndividualComment key={index}>
+                            <p><span>{item.username}</span> {item.text}</p>
+                        </IndividualComment>
+                ))}
+                <Timestamp>{moment(this.state.timestamp, 'MMMM-DD-YYYY hh:mm:ss').fromNow().toUpperCase()}</Timestamp>
```
trevorfehrman: Gj getting moment implemented.

instagram/src/components/LikesContainer/LikesContainer.js
```diff
@@ -0,0 +1,50 @@
...
+class LikesContainer extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            likes: this.props.likes,
+            clicked: false,
+        }
+    }
+
+    changeLikes = () => {
```
trevorfehrman: Again you're keeping track of valuable information here far down the tree.  What if you wanted to use these likes somewhere else in the application (assuming you kept adding features to this project).  Then you'd have to send this data back up to App.js somehow only to pass it down to something else.

PR #650, opened 2018-11-13
Kevin Marker

kmarker1101:


instagram/src/Authentication/Authenticate.js
```diff
@@ -0,0 +1,28 @@
...
+                this.state = {
+                    loggedIn: false
+                };
+            }
+        
+        componentDidMount() {
+            if (!localStorage.getItem('user')) {
+                this.setState({ loggedIn: false });
+            } else {
+                this.setState({ loggedIn: true });
```
johnoro: An interesting alternative to this if else is something like this:
```javascript
this.setState({ loggedIn: !!localStorage.getItem('user') });
```
Here, the first NOT operator (`!`) coerces our null or string into a boolean, and then the next NOT makes it the opposite.

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,97 @@
...
+    }
+    
+    render(){
+        return (
+            <CommentContainer>
+                {this.state.comments.map(comment =>{
+                    return (
+                        <div key={comment.text} className = "commentBody"> 
+                        {/* apparently this needed a unique key */}
+                        <strong>{comment.username}</strong><span className = "commentText">{comment.text}</span>
```
johnoro: Splitting these up onto two separate lines would be better for readability:
```javascript
<strong> {comment.username} </strong>
<span className="commentText"> {comment.text} </span>
```

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,97 @@
...
+            </div>          
+            )
+        })}
+
+        <form onSubmit = {(e) => {this.addNewComment(e)}}>    
+            <div className = "timeStamp">
+                <input className = "commentInput" 
+                    type = 'text' placeholder = " Add a comment... " value = {this.state.newComment}  onChange = {this.textHandler} />
```
johnoro: Here's a better alternative for when you have a lot of props:
```javascript
<input 
  className="commentInput" 
  type="text" 
  placeholder=" Add a comment... " 
  value={this.state.newComment}  
  onChange={this.textHandler} 
/>
```

instagram/src/components/SearchBar/LikeButton.js
```diff
@@ -0,0 +1,26 @@
...
+
+    render(){
+        return (
+            <button  onClick = {this.props.addLike}>
+                <LikeIcon alt="''" src= {LikeButt} />
+            </button>
+    )
+}
+}
+// I know this doesn't need to be a class component, but when I changed it to a functional component, the app broke so I'm leaving it.
```
johnoro: This should work just fine as a functional component; maybe you missed something when changing it?

johnoro (commented):
Good work! Try to make more frequent commits. [Here](https://chris.beams.io/posts/git-commit/) is an article on writing nice commit messages (but you don't necessarily need to be super detailed; all you really need to accomplish is accuracy and frequency).

PR #653, opened 2018-11-13
Trai Compton: PR Initiated

TraiLynne:


instagram/src/components/PostContainer/Posts.js
```diff
@@ -0,0 +1,31 @@
...
+    return(
+        <div className='posts'>
+            {props.posts.map((p, i) => (
+                <Post post={p} key={i}/>
+            ))}
+        </div>
+    )
+}
+
+Posts.propTypes = {
```
szincone: Good job getting your proptypes working. Looks like this is a master prop check, you're matching the data exactly!

instagram/src/components/PostContainer/Post.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Post = props => {
+    return(
+        <p>Post</p>
```
szincone: Good job on setting up your project correctly before adding the tiny details.

PR #655, opened 2018-11-13
Nedim Omerovic - React-Insta-Clone

Civoremo:


instagram/src/App.js
```diff
@@ -0,0 +1,23 @@
...
+  constructor(){
+    super();
+    this.state = {}
+  }  
+
+
+  render() {
+    return (
+      <div className="App">
+        <PostPage />
```
trevorfehrman: You should probably extract the header component out of PostPage and render it here.  There's not a lot of a point to rendering a single master component from what should be your single master component.  Like you could literally save a step here, delete App.js, render PostPage in index.js instead of App.js and nothing about your application would change.

instagram/src/components/PostContainer/PostPage.js
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    handleChange = e => {
+        this.setState({
+            [e.target.name]: e.target.value
+        });
+        console.log(e.target.value)
+    }
+
+    likedToggle = id =>
```
trevorfehrman: This is perfect, awesome job on this

instagram/src/components/PostContainer/PostPage.js
```diff
@@ -0,0 +1,86 @@
...
+            } else return post;
+        })
+    }));
+
+
+    render() {
+
+        let filteredArray = this.state.data.filter(
+            post => {
+                return post.username.indexOf(this.state.searchInputText) !== -1;
```
trevorfehrman: Okay I get this now, this is cool because it scans for any valid substring.  It looks hacky but it's a great solution.   I would suggest using `.includes` for something like this.  It'll make it much more readable.

instagram/src/components/PostContainer/PostPage.js
```diff
@@ -0,0 +1,86 @@
...
+        return (
+        
+            <div className='postPage-container'>
+
+            <div className='searchContainer'>
+            {<SearchBar handleChange={this.handleChange} searchText={this.state.searchInputText} />}
+            </div>
+    
+            <div className='postContainer'>
+            {filteredArray.map( arg => {
```
trevorfehrman: Rather than filtering above and the mapping over the filtered array you could save another step and just use `.filter` here

instagram/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,83 @@
...
+                        <SearchInput 
+                            type="text" 
+                            name='searchInputText'
+                            placeholder='search'
+                            value={props.searchText}
+                            onChange={props.handleChange}
+                        />
+                    </form>
+                </SearchDiv>                    
+                <RightSideIconsDiv >
```
trevorfehrman: Cool I like this organization, gj

instagram/src/components/CommentSection/CommentContainer.js
```diff
@@ -0,0 +1,87 @@
...
+        if(localStorage.getItem(id)) {
+            this.setState({
+                comments: JSON.parse(localStorage.getItem(this.props.postid))
+            });
+        } else {
+            this.setComments();
+        }
+    }
+
+    componentWillUnmount() {
```
trevorfehrman: Excellent use of lifecycle methods

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import './comment.css';
+import PropTypes from 'prop-types';
+
+function CommentSection(props){
+    return (
+        <div className='each-comment-container'>
```
trevorfehrman: Try BEM! 

BEM! BEM! BEM! BEM! BEM! BEM! BEM! BEM!

PR #658, opened 2018-11-13
Yasirah Krueng

ykrueng:


instagram/src/components/PostContainer/PostPage.js
```diff
@@ -0,0 +1,122 @@
...
+
+  render() {
+    const options = {
+      keys: ['username'],
+    };
+    const fuse = new Fuse(this.state.data, options);
+    const filteredData = this.state.searchText === '' ? this.state.data : fuse.search(this.state.searchText);
+    return (
+      <div className="post-page">
+        <SearchBar
```
Kamalnrf: It would have been if you had SearchBar in your App component instead of here.

instagram/src/components/PostContainer/PostPage.js
```diff
@@ -0,0 +1,122 @@
...
+        <div className="posts--container">
+          <div className="posts">
+            {
+              filteredData
+              .map(
+                (post, i) => (
+                  <PostContainer
+                    key={post.username + post.timestamp}
+                    commentText={post.commentText}
+                    submitComment={(c, u, t) => this.submitComment(c, u, t)}
```
Kamalnrf: While reading it is difficult for anyone to understand what `c, u, t` it will be better if you can more expressive parameter names or have comments.

Kamalnrf:
### :heavy_check_mark: What went well:
---
- Commit messages look good 
- Login Page design looks good.

### :reminder_ribbon: Recommendation(s):
---
-  try implementing ability where user can delete only their comments.

PR #659, opened 2018-11-13
Brooks Poltl day 1 MVP

BrooksPoltl:
@synch3


instagram/src/components/Authentication/Authentication.js
```diff
@@ -0,0 +1,51 @@
...
+    class auth extends React.Component {
+        constructor() {
+            super()
+            this.state = {
+                user: '',
+                pass: '',
+                loginStatus: false
+            }
+        }
+        userHandler = event => {
```
llwor94: you can refactor userHandler and passHandler into one method by adding a name property of 'user' and 'pass' to each of your inputs, then create a handleChange method like this: 
```
handleChange = event => {
    this.setState({ [event.target.name]: event.target.value })
}
```

instagram/src/components/PostContainer/Post.js
```diff
@@ -7,7 +7,8 @@ class Post extends React.Component{
     constructor(props){
         super(props);
         this.state={
-            likes: props.post.likes
+            likes: props.post.likes,
```
llwor94: avoid initializing the state with props! This can cause issues down the line with the child component not rerendering upon parent changes in state.

PR #661, opened 2018-11-13
Daniel Mendez React-Insta-Clone

dan2014:


instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,68 @@
+import React, {Component} from 'react';
+import "./CommentSection.css"
+import Comments from './Comments'
+
+
+
+class CommentSection extends Component {
```
trevorfehrman: This functionality and data could be lifted to your App.js and you'd have a single source of truth for all your information.  Again, your solution is valid but I would encourage you to try and lift as much as possible.

instagram/src/components/PostContainer/PostContainer.js
```diff
@@ -0,0 +1,43 @@
...
+
+function PostContainer(props){
+    return(
+        <div className="post-container">
+        {
+            props.post.map( e => (
+                // console.log(e.display)
+                <div className="post" key={e.index} style={{display:`${e.display}`}}> 
+                <Post 
+                    username={e.username}
```
trevorfehrman: You should definitely just send down your entire object `e` here and destructure it from the context of `Post.js`

PR #662, opened 2018-11-13
Jasmine Anderson - MVP

jasmineea:


Synch3:
Hi Jasmine, in our zoom yesterday we talked about how to check to see if we were passing our props down the components through console logs. We went over font-awesome briefly, and we made sure we were identifying components.

Synch3:
Nice progress Jasmine, we discussed passing props in our zoom yesterday and how to identify those props with keys you create as you render components.

PR #663, opened 2018-11-13
chasefulks

chasefulks:


instagram/src/App.js
```diff
@@ -0,0 +1,56 @@
...
+    });
+  }
+   handleInputChange = event => {
+    this.setState({
+      filter: event.target.value
+    });
+  };
+   filterItems = event => {
+    event.preventDefault();
+    if (this.state.filter === "") {
```
Kamalnrf: it would have been better if you had two variables in state one is holding the data and other holds what to display.

instagram/src/components/Auth/Auth.js
```diff
@@ -0,0 +1,55 @@
...
+      event.preventDefault();
+      console.log("works");
+      console.log(this.state.username);
+      console.log(this.state.password);
+      console.log(this.state.loggedIn);
+       localStorage.setItem(
+        this.state.username, JSON.stringify(this.state.password)
+      );
+        
+      this.setState({
```
Kamalnrf: you are not making any changes here?

instagram/src/components/PostContainer/PostPage.js
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import PostContainer from "./PostContainer";
+import SearchBar from "../SearchBar/SearchBar";
+ const PostPage = props => {
+  return (
+    <div>
+      <SearchBar
```
Kamalnrf: SeatchBar component could have been in your App component instead of here

PR #664, opened 2018-11-14
Brandon Campbell

BrandCam:
WIP


instagram/src/components/PostContainer/PostContainer.js
```diff
@@ -0,0 +1,33 @@
...
+import CommentSection from '../CommentSection/CommentSection';
+import './PostContainer.css';
+import * as Icon from 'react-feather';
+
+function PostContainer(props){
+    return(
+        <section className='post'>
+            <div className='post-content'>
+                <div className='post-head'>
+                    <img className='thumbnail' src={props.post.thumbnailUrl}></img><span>{props.post.username}</span>
```
johnoro: These tags ought to be separated for readability:
```javascript
<img className='thumbnail' src={props.post.thumbnailUrl}></img>
<span> {props.post.username} </span>
```

instagram/src/components/PostContainer/PostContainer.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import CommentSection from '../CommentSection/CommentSection';
+import './PostContainer.css';
+import * as Icon from 'react-feather';
+
+function PostContainer(props){
```
johnoro: This is a great candidate for object destructuring. For example:
```javascript
function PostContainer({ post, user }) { 
  // ...
```

PR #666, opened 2018-11-14
Lidiia Gallagher

Lidiia92:


instagram/src/App.js
```diff
@@ -0,0 +1,56 @@
...
+    };
+  }
+
+  handleInputChange = event => {
+    this.setState({
+      filter: event.target.value,
+    });
+  };
+
+  getFilteredPosts = () => {
```
Kamalnrf: I like the way you have thought about it. Looks clever.

instagram/src/components/PostContainer/PostsPage.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+import SearchBar from '../SearchBar/SearchBar';
+import PostList from './PostList';
+
+const PostsPage = props => {
+    return (
+    <React.Fragment>
+        <SearchBar  filter={props.filter} handleInputChange={props.handleInputChange}/>
```
Kamalnrf: It would have been better if the searchbar was in app.js instead of here

instagram/src/components/PostContainer/PostList.js
```diff
@@ -0,0 +1,24 @@
...
+import Post from './Post';
+
+import './PostContainer.css';
+
+
+
+const PostList = props => {
+    let posts = props.postData.map((post, index) => {
+    return <Post key={index} postData={post} 
+    inputComment={props.inputComment}
```
Kamalnrf: formatting seems a bit off here.

instagram/src/components/PostContainer/PostList.js
```diff
@@ -0,0 +1,24 @@
...
+    return <Post key={index} postData={post} 
+    inputComment={props.inputComment} 
+    handleAddComment={props.handleAddComment} 
+    handleChange={props.handleChange} username={props.username}/>; 
+    });
+
+
+  return (
+    <React.Fragment>
+        {posts}
```
Kamalnrf: you could had mapped over the postData in the curly braces also.

PR #667, opened 2018-11-14
Nick Stricker

NickStrick:


instagram/src/App.js
```diff
@@ -0,0 +1,30 @@
...
+  constructor(){
+    super();
+    this.state=({
+      users: [1]
+    })
+  }
+
+  login = () => {
+    localStorage.setItem('username', 'anonymous') 
+    this.setState();
```
johnoro: It looks like you have some artifacts here that you aren't using anymore. Why not get rid of what you don't need? At this point, it looks like `App` could be a functional component, or perhaps could just be removed completely and then have `PostsPage` take its place (still with the `Authenticate` HOC).

instagram/src/Authentication/Authenticate.js
```diff
@@ -0,0 +1,36 @@
...
+        if(!(localStorage.getItem('username') == undefined)){
+            this.setState({
+                loggedIn: true,
+            })
+        }
+    }
+
+      render() {
+        console.log(this.state);
+        if (this.state.loggedIn === false) {
```
johnoro: Why not use the NOT operator? i.e.:
```javascript
if (!this.state.loggedIn)
```

instagram/src/components/CommentSection/Comment.js
```diff
@@ -0,0 +1,24 @@
...
+import './CommentSection.css';
+import PropTypes from 'prop-types';
+
+
+
+const Comment = props => {
+
+    return(
+        <div className='comment'>
+            <p><span>{props.comment.username}</span> {props.comment.text}</p>
```
johnoro: This could probably be broken up for readability, for example:
```javascript
<p>
  <span> {props.comment.username} </span>
  {props.comment.text}
</p>
```

johnoro: I'd also love to see [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring) here.

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,67 @@
...
+        }
+      }
+
+      componentDidMount(){
+        this.setState({ comments: this.props.comments})
+      }
+
+      addNewComment = ev =>{
+        ev.preventDefault();
+        if(!(this.state.inputText === '')){
```
johnoro: Since an empty string is false-y, and any other string is truth-y, you could be tricky and skip the equality operator here.

instagram/src/components/PostContainer/Post.js
```diff
@@ -0,0 +1,41 @@
+import React, { Component } from 'react';
+import './PostContainer.css';
+import PropTypes from 'prop-types';
+
+
+
+
+const Post = props => {
```
johnoro: Imagine getting rid of all of the `props.` down below by doing something as short as this with object destructuring:
```javascript
const Post = ({ addLike, data, id }) => {
  // ...
```

instagram/src/components/CommentSection/Comment.js
```diff
@@ -1,15 +1,21 @@
...
+`;
 
 const Comment = props => {
 
     return(
-        <div className='comment'>
-            <p><span>{props.comment.username}</span> {props.comment.text}</p>
-        </div>
+        <CommentBox >
+            <p><CommentUsername>{props.comment.username}</CommentUsername> {props.comment.text}</p>
```
johnoro: You could probably break this up for better readability:
```javascript
<p>
  <CommentUsername>{props.comment.username}</CommentUsername>
  {props.comment.text}
</p>
```
or, with object destructuring:
```javascript
const { username, text } = props.comment;
return (
// ...
  <p>
    <CommentUsername>{username}</CommentUsername>
    {text}
  </p>
);
```

PR #669, opened 2018-11-14
Grace Kang

bananaboots:


instagram/src/App.js
```diff
@@ -0,0 +1,33 @@
...
+    this.setState({
+      ...this.state,
+      searchQuery
+    })
+  }
+
+  render() {
+    return (
+      <div className="App">
+        <SearchBar onSearch={(searchQuery) => this.onSearch(searchQuery)} />
```
johnoro: I'm curious why you're using an arrow function to wrap this function here. It doesn't seem to serve any purpose besides getting around `this`, so why not pass just the function definition and also make that an arrow function?
```javascript
onSearch = searchQuery => {
  this.setState({
    ...this.state,
    searchQuery
  });
};
// ...
<SearchBar onSearch={this.onSearch} />
```

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,65 @@
...
+            })
+        })
+    }
+
+    render() {
+        return (
+            <Fragment>
+                {this.state.comments.map((comment, index) => <Comment comment={comment} key={index} />)}
+                <p className="timestamp">
+                    {(moment(this.props.timestamp, "MMMM Do YYYY, h:mm:ss a").fromNow()).toUpperCase()}
```
johnoro: This is probably a place where you ought to break this onto other lines for readability. For example:
```javascript
{
  moment(this.props.timestamp, "MMMM Do YYYY, h:mm:ss a")
    .fromNow()
    .toUpperCase()
}
```

instagram/src/components/CommentSection/CommentSection.js
```diff
@@ -0,0 +1,65 @@
...
+                </p>
+                <form onSubmit={this.addComment}>
+                    <input type="text" name="commentText" placeholder="Add a comment..." value={this.state.commentText} onChange={this.changeHandler} />
```
johnoro: The same goes for components with lots of props or elements with lots of attributes! For example:
```javascript
<input 
  type="text" 
  name="commentText" 
  placeholder="Add a comment..." 
  value={this.state.commentText} 
  onChange={this.changeHandler} 
/>
```

instagram/src/components/SearchBar/SearchBar.js
```diff
@@ -18,19 +72,21 @@ class SearchBar extends Component {
...
                     <i className="fab fa-instagram"></i>
-                    <span>|</span>
-                    <img src="instagram_logo.svg" />
-                </div>
+                    <Line />
+                    <img className="logo-img" src="instagram_logo.svg" alt="logo" />
+                </LogoDiv>
                 
-                <input className="searchbar" type="text" placeholder="Search" value={this.state.searchQuery} onChange={this.changeHandler}></input>
+                <SearchbarInput type="text" placeholder="Search" value={this.state.searchQuery} onChange={this.changeHandler}></SearchbarInput>
```
johnoro: This should be broken up for readability (and especially for avoiding horizontal scroll):
```javascript
<SearchbarInput 
  type="text" 
  placeholder="Search" 
  value={this.state.searchQuery} 
  onChange={this.changeHandler}
>
</SearchbarInput>
```

instagram/src/components/PostContainer/PostPage.js
```diff
@@ -0,0 +1,33 @@
...
+    this.setState({
+      ...this.state,
+      searchQuery
+    })
+  }
+
+  render() {
+    return (
+      <div className="App">
+        <SearchBar logOut={this.props.logOut} onSearch={(searchQuery) => this.onSearch(searchQuery)} />
```
johnoro: Why make this a bit roundabout and have the arrow function here? It'd be easier and arguably better to have the method itself be an arrow func:
```javascript
  onSearch = searchQuery => {
    this.setState({
      ...this.state,
      searchQuery
    })
  }
```

instagram/src/components/PostContainer/Post.js
```diff
@@ -22,39 +72,39 @@ class Post extends Component {
...
+                    likes: prevState.likes + 1
+                };
             })
         }
         else if (event.target.classList.contains('fas')) {
-            this.setState({
-                likes: this.props.post.likes
+            this.setState((prevState) => {
+                return {
+                    likes: prevState.likes - 1
```
johnoro: This `if...else if` chain looks a little unnecessary. Could you not just use an `if...else` since you're checking the opposite of what the `else if` is checking in your `if`? For example:
```javascript
if (!event.target.classList.contains('fas')) {
  this.setState(prevState => {
    return {
      likes: prevState.likes + 1;
    };
  });
} else {
  this.setState(prevState => {
    return {
      likes: prevState.likes - 1;
    };
  });
}
```
I also see a bit of repetition here. Why not use a variable in your if else instead, and then `setState` afterwards? For example:
```javascript
let amount;
if (!event.target.classList.contains('fas')) {
  amount = 1;
} else {
  amount = -1;
}
this.setState(prevState => {
  return {
    likes: prevState.likes + amount;
  };
});
```
You could even use a ternary now to shorten your code:
```javascript
const { classList } = event.target;
const amount = classList.contains('fas') ? -1 : 1;
this.setState(prevState => {
  return {
    likes: prevState.likes + amount;
  };
});
```

instagram/src/components/Login/Login.js
```diff
@@ -0,0 +1,63 @@
...
+        })
+    }
+
+    changeHandler = (event) => {
+        this.setState({
+            ...this.state,
+            [event.target.name]: event.target.value
+        }, () => {
+            if (this.state.usernameText !== '') {
+                document.querySelector('.loginbutton').style.backgroundColor = '#5097E9';
```
johnoro: _oh no, not a querySelector!_ 
It's better, at least slightly, to use [refs](https://reactjs.org/docs/refs-and-the-dom.html) if you need to do something like this to ensure that you're getting the right thing. It'd look like [this](https://reactjs.org/docs/refs-and-the-dom.html#creating-refs). Try it out!

instagram/src/components/Authentication/Authenticate.js
```diff
@@ -0,0 +1,39 @@
...
+          }
+
+        render() {
+            return (window.localStorage.getItem('username')) ?
+                (
+                    <WrappedComponent logOut={this.logOut} />
+                ) :
+                (
+                    <Login />
+                )
```
johnoro: I'm not sure if you really need the parentheses here.

instagram/src/App.js
```diff
@@ -1,31 +1,18 @@
...
-    })
-  }
 
   render() {
     return (
-      <div className="App">
-        <SearchBar onSearch={(searchQuery) => this.onSearch(searchQuery)} />
-        <PostContainer searchQuery={this.state.searchQuery} posts={this.state.posts} />
-      </div>
+      <PostPage />
```
johnoro: Your App seems to just be rendering your PostsPage now. You could probably get rid of your App component easily without any loss of functionality or organization.

PR #670, opened 2018-11-14
Tommy Huynh

tommaay:


instagram/src/components/PostContainer/PostsPage.js
```diff
@@ -0,0 +1,19 @@
+import React, { Component } from "react";
+import PostContainer from "./PostContainer";
+import SearchBar from "../SearchBar/SearchBar";
+
+const PostsPage = props => {
+  return (
+    <>
+      <SearchBar
```
Kamalnrf: You could have had your SearchBar in app.js instead of here because your search bar is in nav.

Kamalnrf (commented):
### :heavy_check_mark: What went well:
---
- The way you have separated posts looks 
- MVP looks good
- I also like the way you have used styled-components 

### :reminder_ribbon: Recommendation(s):
---
- Check how you can pass props in styled components 
- Check react docs on thinking in react

PR #671, opened 2018-11-14
Trai Compton

TraiLynne:


instagram/src/components/PostContainer/Post.js
```diff
@@ -0,0 +1,118 @@
...
+                        <CardHeader>
+                            <img src={this.post.thumbnailUrl} alt='User Thumbnail'/>
+                            <p>{this.post.username}</p>
+                        </CardHeader>
+                        <CardBody>
+                            <img src={this.post.imageUrl} alt='Post'/>
+                        </CardBody>
+                        <CardFooter>
+                            Posted: {this.post.timestamp}
+                            <Button color="danger" onClick={this.toggleComments}><i className="far fa-comment"> {this.state.comments.length}</i></Button>{' '}
```
trevorfehrman: Cool, usin' font awesome.

instagram/src/components/PostContainer/PostsPage.js
```diff
@@ -0,0 +1,57 @@
...
+        }
+    }
+
+    componentDidMount() {
+        this.setState({
+            posts: dummyData
+        })
+    }
+
+    handleChange = e => {
```
trevorfehrman: Yeah this is good.  Notice how this pattern lets you control a child component from here.  If you were to change `searchText` on your state manually it would update the value of the input field.  This is called 2-way-binding.

PR #672, opened 2018-11-15
vanilla search bar done

alexcartaz:
vanilla search bar and navbar created. next is post cards and main feed. realized that my solution for making search icon appear in placeholder text is janky and that I'll need to replace it by creating a fix-position div that disappears onclick in form -- or smtg along those lines


instagram/src/components/PostContainer/ImageThumbnail.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import './Header.css';
+
+function HeaderImage(props) {
+  return (
+    <React.Fragment>
+    	<img src={props.imageUrl} className="header-image"></img>
+    </React.Fragment>
```
johnoro: Why have a fragment here? It seems like it's not serving any purpose!

instagram/src/components/PostContainer/PostCard.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+function PostCard(props) {
+  return (
+    <div className="postCardContainer"></div>
```
johnoro: For CSS class names, you should keep them to be more in line with what you expect from CSS (i.e. `post-card-container`).

instagram/src/components/PostContainer/CardBanner.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import './Card.css';
+
+function CardBanner(props) {
+  return (
+    <React.Fragment>
+    	<img src={props.imageUrl} className="card-banner"></img>
+    </React.Fragment>
```
johnoro: Why use a fragment here?

johnoro (commented):
Good work! Try to commit more than once for this much work. That sort of thing probably wouldn't fly in a team environment.

alexcartaz:
running list of issues from insta clone project:
* couldnt figure out how to place a space between username and comment text in <Comment /> jsx
* couldn’t figure out how to correctly install, import, and use moment.js (though I was able to do correct date conversion just vanilla js Date)
* when passing a prop from a parent to a child via a function (to cleanse or modify data), it seemed to break the PropType check (see Comment Footer / warning)
* Need to figure out how to add Font Awesome Icons as  border-only icons (not-filled)
* Need to figure out how to properly overlay div for font-awesome search icon for search box display
* Still need to figure out prettier text formatting
* Did not get all of search working perfectly: backspace/delete don’t work. No way to unfiltered without refreshing the page.

alexcartaz:
finished day 3 vanilla. added some timestamping functionality for the log in page local storage so that it didn't keep the user flagged as logged forever once it had happened once.

PR #675, opened 2018-11-26
Joseph Eastman

josepheastman:


instagram/src/components/PostContainer/Post.js
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+import CommentSectionContainer from "../CommentSection/CommentSectionContainer";
+import PostHeader from "./PostHeader";
+import "./Post.css";
+
+const Post = props => {
+  return (
+    <div className="post-border">
```
ashwins93: Instead of setting fixed width on the comments container set the width on this `.post-border` class so that the whole card has the same width.

```css
.post-border {
  width: 640px;
  margin: 0 auto;
}
```

PR #677, opened 2018-11-27
David Situ 

Codechiha:
Search bar placed
PostContainer
Comment pending


ashwins93 (commented):
The cards look good. You can use font awesome icons for the navbar. Referring solution code when stuck is a good thing but please refrain from copying the code fully.

PR #691, opened 2018-12-13
My PR

BoxOfCereal:
Hello 👋 @theBrockstar


instaclone/src/Authentication/Authenticate.js
```diff
@@ -0,0 +1,45 @@
...
+            this.setState({loggedIn: true})
+            this.setState({user:myStorage.getObject('user')})
+        }
+        
+    }
+
+     // updates local storage with the username and also sets the state of apps this.state.user
+    login = (username, password) => { 
+        myStorage.setObject('user', {username:username});
+        this.setState({user:  myStorage.getObject('user').username})
```
TheBrockstar: This is great! However, how might you refactor so that instead of just storing a username, it checks the username against an object of users and passwords to see if the user exists and if the password is correct and then logs the user in if it is?

PR #696, opened 2018-12-13
Daniel Lim

dlimla:
Day 1 

Needed to take care of something so I'm a little behind didn't get much done but I'm working on it


Lndubose:
# Day 1
## Tasks (Day I)
- [ ] The CommentSection component will receive the array of comments as props and render a Comment component with the username of the poster as well as the post's text. [ ]Additionally, there should be an input box that allows users to submit a new comment for any post. We'll work on posting new comments tomorrow.
Be sure to check the Types of the data you are passing around as props in the components that will be using props to present data as DOM elements. This should be linked to your Comment component that Comment Section will render and potentially to your Post component that Post Container will render.
You are free to leverage the Bootstrap library for this project for the purposes of theming and styling. I recommend the awesome reactstrap library, which is a library of Bootstrap components that have been implemented using React, so they're really easy to just drop straight into React projects.
In addition to Bootstrap for theming, you'll want to add your own styles via CSS. To keep things organized, have the CSS file that corresponds with a component live in the same directory as the component file.

samsisle:
#### Tasks (Day III)

- Create a `<PostsPage />` component in your `components/PostsContainer` directory.

  - [X] You'll have to move a lot of what is rendered in `app.js` to this new component
  - [X] In app.js, render the `PostsPage` component.
  - [X] Make sure the app working as it was before since it has been re-factored now.
  - [X] This is to ensure that we clean up our App component a little bit before we re-factor it to be wrapped up in an HOC

  - [X] Create a directory called `Authentication`
  - [X] Inside that directory create a component called `<Authenticate />`. This is where all of the magic sis going to happen.
  - [X] This component should be able to take in a component as an argument, and it will return a `class` component.
  - [X] Inside of `<Authenticate />'s` render method, you'll want to return the Component that gets passed into it.
  - [X] Be sure to export out this component.
  - [X] Head over to App.js and `import` in our new `Authenticate` Higher Order Component and pass in `App`.

- Build out the LoginPage component. You can design it how you like

  - [X] In your `components` directory, create a directory called `Login` and add a new file called `Login.js`.
  - [ ] There should be a `username` input, a `password` input, and a `Login` button.
  - [X] The component should invoke the `login` function in `app.js` when a user logs in.
  - [ ] This login function should set a `username` on `localStorage`. You'll need to check local storage to see if a user is logged in.
  - [ ] Be sure to force the page to reload when a user logs in so that our component un-mounts and mounts again.

- Extending the functionality of the HOC to conditionally render the `LoginPage` or the `App`

  - [X] Inside of `Authenticate` we need to add a constructor to hold our state data.
  - [X] On state we need a user `loggedIn` boolean flag.
  - [ ] On `componentDidMount` we need to check `localStorage` to see if a user is logged in.
  - [X]  Inside of the render function we will check `if a user is logged in`
  - [X] If a user is logged in we will return the `<App />`, else we will return the `<LoginPage>`

#### Notes
If you have any questions, make sure to ask after 20 minutes if you're still stuck! Keep it up man

#### Rating ⭐

PR #700, opened 2018-12-13
React Insta Clone Zach irvin

badCompany55:


Lndubose:
# Day 1
## Tasks (Day I)
- [ ] The CommentSection component will receive the array of comments as props and render a Comment component with the username of the poster as well as the post's text. 
- [ ] Additionally, there should be an input box that allows users to submit a new comment for any post. We'll work on posting new comments tomorrow.

## Stretch Problems (Day I)

# Day 2
## Tasks (Day II)
### Use lifecycle methods
  - [ ] In `app.js` use `componentDidMount()` to set your data to the component's state. Use the state object to pass data to your `PostContainer` component.
### Adding comments
  - [ ] Lets divide up the data that we've been working with this far by separating the comments array onto a new component's state. Pass down the comments through each post to the CommentSection component. Be sure that you set the incoming `comments` props data on the state of the CommentSection component.
  - [ ] Create a function in `CommentSection.js` called `addNewComment` that takes in an event and an index number. The function will add the comment that is on the event object to the post that is at that index number.
  - [ ] Pass the `addNewComment` function down the component tree to where ever you have your 'Add a comment...' input.
  - [ ] The 'Add a comment...' input should be wrapped in a `<form></form>` element. Using that form's `onSubmit` event handler, invoke the `addNewComment` function and pass it the required arguments it needs to add a new comment.
  - [ ] Update your state with the new comment (This should trigger your component tree to "re-render" with your new comment on that post).
### Liking posts
  - [ ] This will be set up a lot like the 'Add a comment...' input. Pass a function down to where your heart icon is, and use `onClick` event handler to increment that post's likes.
### Search
  - [ ] Set up the search bar will like the comment input and the like button. In your function, filter out any post whose username doesn't match the search term passed in, then update the state with the resulting data.

## Stretch Problems (Day II)

- [ ] Persist your data using `localStorage`. If done correctly, you will be able to refresh your page and still see your new comments.
- [ ] Add the ability to delete a comment from your data. If your data is in `localhost`, make sure to delete it from there as well.


### I am not able to see your instagram work through your PR. I was able to find it through your repositories but as of right now, it is not showing in your PR. I graded based off what I found in your repo.

## Commits
- Commits look good. Very descriptive. Could look into headers and bodies for commits to make your commit game even stronger.

## Improvements
- Saw you were trying to mess with the localStorage but something was not rendering your posts for me when I tried to upload it.

## Rating
3

Overall it looks really good. I graded the Sprint based on your `modified the addcommentFunction...` commit. The one after this would not show the posts for me. I think it has something to do with the componentDidMount in App.js. Otherwise it looks good, keep up the good work.

PR #709, opened 2018-12-19
Instagram Clone

LeeTann:
used CRA to boilerplate app and worked on search bar


keirankozlowski:
### Objective
After two days of working on this project, you should be able to:

- Demonstrate the the ability to use create-react-app to boilerplate a react application
- Describe and be able to use PropTypes to 'type check' specific data being passed down to a child component
- Demonstrate the ability to use defaultProps in a React component
- Understand the concept of React component lifecycles, along with the major lifecycle methods such as render, componentDidMount.

### MVP
**Objectives**

- [x] Search bar renders
- [x] Search bar properly searches based on username
- [x] Post / Comment section renders for each datapoint
- [x] When a comment is added, there is a username for it
- [x] New comments can be added
- [x] Proper use of PropTypes

### Great
- Everything is rendered in the proper order.
- Your rudimentary styling is good for the stage you're in.
- Great job getting the search and add comment functionality working!
- Nice job with passing the props and setting state properly.
- Your code is easy to read and well-organized.

### Requests for Improvement
- You have a typo for insta-comment.PNG calling it insta-comment.png which is causing it to fail; it looked fine once I fixed this. Please update this immediately.
- You could use some more PropTypes, but what you have is enough for MVP.
- If you try to search more than once, there is a user not found error.
- If you press enter on a blank comment, it will add a blank comment.
- Work on the styling.

### Commits
- Decent job with commit number; still could use improvement on the actual messages.

### Rating: 2

samsisle:
#### Tasks (Day III)

- Create a `<PostsPage />` component in your `components/PostsContainer` directory.

  - [X] You'll have to move a lot of what is rendered in `app.js` to this new component
  - [X] In app.js, render the `PostsPage` component.
  - [X] Make sure the app working as it was before since it has been re-factored now.
  - [X] This is to ensure that we clean up our App component a little bit before we re-factor it to be wrapped up in an HOC

  - [X] Create a directory called `Authentication`
  - [X] Inside that directory create a component called `<Authenticate />`. This is where all of the magic sis going to happen.
  - [X] This component should be able to take in a component as an argument, and it will return a `class` component.
  - [X] Inside of `<Authenticate />'s` render method, you'll want to return the Component that gets passed into it.
  - [X] Be sure to export out this component.
  - [X] Head over to App.js and `import` in our new `Authenticate` Higher Order Component and pass in `App`.

- Build out the LoginPage component. You can design it how you like

  - [X] In your `components` directory, create a directory called `Login` and add a new file called `Login.js`.
  - [X] There should be a `username` input, a `password` input, and a `Login` button.
  - [X] The component should invoke the `login` function in `app.js` when a user logs in.
  - [X] This login function should set a `username` on `localStorage`. You'll need to check local storage to see if a user is logged in.
  - [X] Be sure to force the page to reload when a user logs in so that our component un-mounts and mounts again.

- Extending the functionality of the HOC to conditionally render the `LoginPage` or the `App`

  - [X] Inside of `Authenticate` we need to add a constructor to hold our state data.
  - [X] On state we need a user `loggedIn` boolean flag.
  - [X] On `componentDidMount` we need to check `localStorage` to see if a user is logged in.
  - [X]  Inside of the render function we will check `if a user is logged in`
  - [X] If a user is logged in we will return the `<App />`, else we will return the `<LoginPage>`

#### Comments
Great job!

#### Rating ⭐⭐

PR #710, opened 2018-12-27
Karla Parraga

karlamparraga:
Tasks (Day I)

The root App component of your application should import the dummy data from the dummy-data.js file with import dummyData from './dummy-data'; and iterate over said data, passing each individual object as a prop to an instance of PostContainer.

Each PostContainer component will then pass the array of comments on each post object as a prop to an instance of the CommentSection component.

The CommentSection component will receive the array of comments as props and render a Comment component with the username of the poster as well as the post's text. Additionally, there should be an input box that allows users to submit a new comment for any post. We'll work on posting new comments tomorrow.

Be sure to check the Types of the data you are passing around as props in the components that will be using props to present data as DOM elements. 

This should be linked to your Comment component that Comment Section will render and potentially to your Post component that Post Container will render.


kevin-sooter:
Good job so far!  Just keep doing what you can.

abravebee:
## Objectives

# DAY ONE
## MVP
- [x] Static Search Bar
- [x] import dummy data
- [x] `Post` component
- [x] `CommentSection` component
- [x] `Comment` component
- [x] PropTypes
- [x] Styles

## Stretch
- [x] Add a comment
- [ ] moment.js

## Great
**Rating and commentary based on commits made by Day One due date: 1/16/2019**
- Great progress! It looks like you're becoming more comfortable with passing data in React. Nice use of console.logs for debugging and understanding as well!

## Rating: 3

# DAY TWO
## MVP
### Lifecycle Methods
- [x] `componentDidMount()` in `app.js`
- [x] Pass data to `PostContainer`
### Adding comments
- [x] Pass comments to `CommentSection`
- [x] `addNewComment` function
- [x] Pass `addNewComment` to comment input
- [ ] Use `onSubmit` to invoke `addNewComment`
### Liking posts
- [x] Increment using `onClick` event handler
### Search
- [x] Function to filter out posts by username

## Stretch
- [ ] Persist data using `localStorage`
- [ ] Delete comments
- [ ] Fuzzy search

## Good
- Nice work meeting MVP! Your app functionality is smooth and I'm able to parse your code quite well. Good job!

## Requests for Improvement
- Since you're using the variable `search` on state as your input handler, it might help readability for future developers to use a different variable inside your filterUsernames function, like `filteredPosts` or `searchResults`.
- I'd also recommend either removing large pieces of commented out code if you no longer need them, or annotating them for future study if there's a reason you want to be able to still refer to them.
- While your `changeComment` function certainly works as intended, I recommend looking into `onSubmit` and considering how this might allow you to cut down on the amount of logic this function contains.

## Additional Questions/Comments
- What problem is the inclusion of `onKeyDown` solving? If I'm not mistaken, you should be able to use just `onChange`, but given your comments during stand-up, I'm curious to know if there was something you ran into that `onKeyDown` fixed.

## Rating: 2

PR #731, opened 2019-01-17
Re-Doing Insta-Clone

mustafainaya:


abravebee:
## Objectives

# DAY ONE
## MVP
- [x] Static Search Bar
- [x] import dummy data
- [x] `Post` component
- [x] `CommentSection` component
- [x] `Comment` component
- [x] PropTypes
- [ ] Styles

## Stretch
- [ ] Add a comment
- [ ] moment.js

## Good
- Good work building bite-sized components and passing data down through props! You're just missing styles but we discussed that you're saving them for use with styled-components.

## Requests for Improvement
- There is a typo in the ternary operator you're using to define props for the PostContainer, where both conditions call for `this.state.filteredPosts` to be sent down, so we're not able to see any posts until we search for something.
- There is also a typo in your PostHeader component: `props.thumbaniUrl` instead of `props.thumbnailUrl`. Make sure to always double check that your page is displaying all the data it should and that you don't have any typos when referencing props.
- You're sending `this.state.instaData` down on props, but there is nothing called `instaData` on state. Make sure to remove old variable references when they're no longer needed.
- There are a couple of warnings in your console; how can those be resolved?

## Rating: 2

# DAY TWO
## MVP
### Lifecycle Methods
- [x] `componentDidMount()` in `app.js`
- [x] Pass data to `PostContainer`
### Adding comments
- [x] Pass comments to `CommentSection`
- [ ] `addNewComment` function
- [x] Pass `addNewComment` to comment input
- [ ] Use `onSubmit` to invoke `addNewComment`
### Liking posts
- [x] Increment using `onClick` event handler
### Search
- [x] Function to filter out posts by username

## Stretch
- [ ] Persist data using `localStorage`
- [ ] Delete comments
- [ ] Fuzzy search

## Needs Improvement
- Your search functionality works great and you're using your lifecycles correctly! However, it looks like the comment functionality isn't quite working yet. I don't see any hearts on the individual posts to test your Like increment functionality and clicking the `div` you have your onClick attached to unfortunately doesn't work.

## Requests for Improvement
- There is a typo in your `CommentInput` component where you're referencing `props.submitComment` but in the `CommentSection` component, you've sent the function down as `commentSubmit`. Fixing this typo leads to another debugging challenge as some type error warnings pop up in the console.
- Think about your `setComments()` function. We want to use this to store our comments new ones are added, and be able to pull that updated data from it. Currently you are using it in `componentWillMount()` (which is correct), but what other lifecycle method have used for populating state? And how can we utilize it in `commentSubmit` to make sure state is always being populated with the newest array of comments?
- Moving your onClick to a button for the like increment does show that the functionality works, but make sure you have a heart or something more clickable than a div to attach it to.

## Rating: 1

PR #733, opened 2019-01-19
Paul DeYonghe's React-Insta-Clone

pdeyo70:


abravebee:
## Objectives

# DAY ONE
## MVP
- [x] Static Search Bar
- [x] import dummy data
- [ ] `Post` component
- [ ] `CommentSection` component
- [ ] `Comment` component
- [ ] PropTypes
- [ ] Styles

## Stretch
- [ ] Add a comment
- [ ] moment.js

## Needs Improvement
- You're importing your data and sending it down as props from `App.js` correctly, but haven't yet properly implemented map in your container components.

## Requests for Improvement
- In PostContainer, you'll need a `map` so you can create a Post component for each post in the props object. Remember that the props object that's been sent down to PostContainer currently has all the posts on it, so we can't grab a single element like `props.posts.username` quite yet.
  - Tip: you can add a `console.log` to your component above the `return` statement so you can see what's been sent down. Example:
```
const PostContainer = props => {
console.log("props.posts in PostContainer", props.posts)
return (
<div>
</div>
);
};
```
- This will help you see in the console what the data that you're passing down looks like and it's super useful for debugging! If anything in your app is ever behaving in a way that's unexpectedly, throw around a bunch of console logs. It will help you follow the flow of the app and visualize the data that you're working with in each component.
- I would also recommend putting console logs into model solutions so you can compare your own data flow with the solution.

## Questions/Comments
- Your commits say you've moved on to day 2 but it looks like you're still missing some of the day 1 MVP. Remember the 20-minute rule! I'm here to help if you're stuck. We can go over some of these concepts during class tonight if you'd like!

## Rating: 1

PR #815, opened 2019-06-12
fix the readme so it's clearer

Ladrillo:


DAY_TWO_README.md
```diff
@@ -26,10 +26,10 @@
...
-  - In `app.js` use `componentDidMount()` to set your data to the component's state. Use the state object to pass data to your `PostContainer` component.
+  - In `App.js` use `componentDidMount()` to set your data to the component's state. This dummy data held in state is an array you can map over, and output a `PostContainer` for each element in the array. Make sure to inject the right props into each `PostContainer` component.
 - Adding comments
   - Lets divide up the data that we've been working with this far by separating the comments array onto a new component's state. Pass down the comments through each post to the CommentSection component. Be sure that you set the incoming `comments` props data on the state of the CommentSection component.
-  - Create a function in `CommentSection.js` called `addNewComment` that takes in an event and an index number. The function will add the comment that is on the event object to the post that is at that index number.
+  - Create a function in `App.js` called `addNewComment` that takes in an event and an index number. The function will add the comment that is on the event object to the post that is at that index number.
```
AlexEntrepreneur: I think using `id` instead of `index` is more clear. 👍

Ladrillo: Actually you're so right

